PR #20, opened 2016-03-29
Configure eclipse to not remove trailing whitespace on empty lines

kaklakariada:
This avoids breaking the license header check which expects trailing whitespace on empty lines in comments.
Note: This setting only seems to affect the license header. For all other places this has no negative effects.


redcatbear:
Not exactly my favourite solution. But it will do for the moment.

Issue #34, opened 2016-04-30
Specobject importer: skip XML files with unknown root element


kaklakariada:
Avoid flooding the log with warning caused by unknown XML elements by using a white list of root element names.

PR #35, opened 2016-04-30
#34 Whitelist XML root elements

kaklakariada:
Skip files with unknown root element. See #34

### DIFF
--- a/src/main/java/openfasttrack/importer/specobject/ImportHelper.java
+++ b/src/main/java/openfasttrack/importer/specobject/ImportHelper.java
@@ -67,7 +67,7 @@ class ImportHelper
 
             if (currentEvent.isStartElement())
             {
-                if (skipFile(currentEvent.asStartElement()))
+                if (isXmlRootWhitelisted(currentEvent.asStartElement()))
                 {
                     return;
                 }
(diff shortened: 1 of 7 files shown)


kaklakariada:
Codecov/changes fails with status "14 files have unexpected coverage changes not found in diff."
No idea what this means, I would ignore it.

src/main/java/openfasttrack/importer/specobject/ImportHelper.java
```diff
@@ -55,36 +67,73 @@ void runImport() throws XMLStreamException
...
+                if (skipFile(currentEvent.asStartElement()))
+                {
+                    return;
+                }
                 foundStartElement(currentEvent.asStartElement());
             }
         }
     }
 
+    private boolean skipFile(final StartElement startElement)
```
redcatbear: The name "skipFile" does not fit. No file is skipped in this method. Instead this method checks the conditions that lead to skipping a file in another method.

Suggested: "isXmlStartWhitelisted()"

src/main/java/openfasttrack/importer/specobject/ImportHelper.java
```diff
@@ -55,36 +67,73 @@ void runImport() throws XMLStreamException
...
+    {
+        if (this.validRootElementFound)
+        {
+            return false;
+        }
+        if (rootElementWhitelisted(startElement))
```
redcatbear: verb missing in method name.

src/main/java/openfasttrack/importer/specobject/ImportHelper.java
```diff
@@ -55,36 +67,73 @@ void runImport() throws XMLStreamException
...
+
+    private boolean rootElementWhitelisted(final StartElement rootElement)
+    {
+        final String elementName = rootElement.getName().getLocalPart();
+        return SPECDOCUMENT_ROOT_ELEMENT_NAME.equals(elementName);
+    }
+
     private void foundStartElement(final StartElement element) throws XMLStreamException
     {
         switch (element.getName().getLocalPart())
```
redcatbear: Code duplication with line 97. Better pass around element name instead?

redcatbear:
Only small findings. Can be merged after correction.

src/main/java/openfasttrack/importer/specobject/ImportHelper.java
kaklakariada: We need the StartElement later in `foundStartElement()` to get attributes from it. I would prefer not to add an parameter for the element name with duplicate information.

kaklakariada:
Thank you for the review, I will merge the PR.

Issue #30, opened 2016-04-13
Issue count per specification item in plain text report


redcatbear:
As a software developer
I want to see the number of broken links and duplicates per specification item
in order to fix the issues faster.
# Acceptance Criteria

Per specification item
1. ... the number of broken incoming links is displayed
2. ... the number of broken outgoing links is displayed
3. ... the number of duplicates is displayed

redcatbear:
Done already.

PR #38, opened 2017-01-15
#30 count per item

redcatbear:


src/main/java/openfasttrack/core/LinkedSpecificationItem.java
```diff
@@ -157,6 +157,30 @@ public void addCoveredArtifactType(final String artifactType)
...
+     * Get a list of all artifact types that have unwanted coverage.
+     * 
+     * @return list of over-covered artifact types.
+     */
+    public List<String> getOverCoveredArtifactTypes()
+    {
+        final List<String> overCovered = new ArrayList<>(getCoveredArtifactTypes());
+        overCovered.removeAll(getNeedsArtifactTypes());
+        return overCovered;
+    }
```
redcatbear: Here the streams variant would have been harder to read. Due to the low number of items probably also slower. Will be replaced by a bit field in performance enhanced version anyway.

PR #41, opened 2017-01-15
Fix sonarqube issues

kaklakariada:
Fixed some sonar findings from https://sonarqube.com/component_issues/index?id=com.github.kaklakariada%3Aopenfasttrack#resolved=false


src/main/java/openfasttrack/core/LinkedSpecificationItem.java
```diff
@@ -200,7 +200,7 @@ public boolean isCoveredDeeply()
         boolean covered = isCoveredShallow();
         for (final LinkedSpecificationItem coveringItem : getLinksByStatus(LinkStatus.COVERS))
         {
-            covered = (covered && coveringItem.isCoveredDeeply());
+            covered = covered && coveringItem.isCoveredDeeply();
```
redcatbear: Out of curiosity: Sonar complains about "superfluous" parentheses?

src/main/java/openfasttrack/core/LinkedSpecificationItem.java
```diff
@@ -236,9 +236,7 @@ public boolean isDefect()
      */
     public int countOutgoingLinks()
     {
-        return countLinksWithPredicate((entry) -> {
-            return entry.getKey().isOutgoing();
-        });
+        return countLinksWithPredicate(entry -> entry.getKey().isOutgoing());
```
redcatbear: Yep, that is more readable.

src/main/java/openfasttrack/importer/markdown/MdPattern.java
```diff
@@ -66,6 +66,11 @@ public Pattern getPattern()
 
     private static class PatternConstants
     {
+        private PatternConstants()
```
redcatbear: :+1:

src/main/java/openfasttrack/importer/markdown/Transition.java
```diff
@@ -27,16 +27,15 @@
     private final State from;
     private final State to;
     private final MdPattern markdownPattern;
-    private final TransitionAction transition;
+    private final TransitionAction transitionAction;
```
redcatbear: Did Sonar complain here?

src/main/java/openfasttrack/importer/specobject/ImportHelper.java
```diff
@@ -105,46 +105,51 @@ private void foundStartElement(final StartElement element) throws XMLStreamExcep
         switch (element.getName().getLocalPart())
         {
         case SPECDOCUMENT_ROOT_ELEMENT_NAME:
-            LOG.finest("Found XML root element '" + SPECDOCUMENT_ROOT_ELEMENT_NAME + "' at "
+            LOG.finest(() -> "Found XML root element '" + SPECDOCUMENT_ROOT_ELEMENT_NAME + "' at "
```
redcatbear: Is the old interface deprecated?
I found the other non-lambda version nicer to read.

And (without prove) I also suspect it to be faster.

src/main/java/openfasttrack/importer/specobject/SingleSpecobjectImportHelper.java
```diff
@@ -69,20 +69,26 @@ void runImport() throws XMLStreamException
                 foundStartElement(currentEvent.asStartElement());
                 break;
             case XMLStreamConstants.END_ELEMENT:
-                if (currentEvent.asEndElement().getName().getLocalPart().equals("specobject"))
+                if ("specobject".equals(currentEvent.asEndElement().getName().getLocalPart()))
```
redcatbear: Safer?
Because it contradicts the natural reading order.

src/main/java/openfasttrack/importer/markdown/Transition.java
kaklakariada: The field has the same name as the class, Sonar says this is confusing.

src/main/java/openfasttrack/importer/specobject/ImportHelper.java
kaklakariada: It is faster like this, because the String is not built when the log level is not active.

src/main/java/openfasttrack/importer/specobject/SingleSpecobjectImportHelper.java
kaklakariada: This avoids nullpointer exceptions

PR #42, opened 2017-01-16
#30: GNU-convention-compliant command line parsing.

redcatbear:
This branch changes the command line arguments to GNU-convention. It also supports chained single-character arguments (like in "tar -czvf foo.tar.gz").

Coverage went down despite my best efforts since Cobertura does not count code with exceptions, even if you have tests to expect them. :disappointed:


src/main/java/openfasttrack/cli/CommandLineInterpreter.java
```diff
@@ -116,10 +121,30 @@ public void parse()
         }
     }
 
+    private void handleSingleCharacterArgument(final ListIterator<String> iterator,
+            final String argument)
+    {
+        final String arguments = argument.replaceFirst(SINGLE_CHAR_ARG_PREFIX, "").toLowerCase();
+
+        int position = 0;
+        for (final String character : arguments.split(""))
```
kaklakariada: This splits a string into chars? This is not obvious. Maybe extract this to method `splitIntoCharacters()`? Or use [String.charAt(i)](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#charAt-int-) as you need the position counter anyway?

src/test/java/openfasttrack/cli/TestCommandLineInterpreter.java
```diff
@@ -147,6 +168,26 @@ public void testCombinedParamters()
         assertThat(stub.getC(), equalTo(StubEnum.VALUE2));
     }
 
+    @Test
+    public void testChainedSingleCharacterParameters()
+    {
+        final CommandLineArgumentsStub stub = parseArguments("-bda", "value_a", "value_1",
+                "value_2");
+        assertThat(stub.getUnnamedValues(), equalTo(asList("value_1", "value_2")));
```
kaklakariada: Maybe use Matcher `containsInOrder()`?

src/test/java/openfasttrack/cli/TestCommandLineInterpreter.java
```diff
@@ -147,6 +168,26 @@ public void testCombinedParamters()
...
+        assertThat(stub.getA(), equalTo("value_a"));
+    }
+
+    @Test
+    public void testChainedSingleCharacterParametersMustFailIfNonBooleanMisplaced()
+    {
+        final CommandLineArgumentsStub stub = new CommandLineArgumentsStub();
+        expectParseException(stub, asList("-bad", "value_a"), "No value for argument 'a'");
+        assertThat(stub.isB(), equalTo(true));
+        assertThat(stub.isD(), equalTo(true));
```
kaklakariada: These two assertions will never be executed as `expectParseException()` does not catch the exception. The exception is handled by the junit rule.

src/main/java/openfasttrack/cli/CommandLineInterpreter.java
redcatbear: Rewrote with regular counting loop. Used "substring" since I need Strings not chars to reuse existing functions.

src/test/java/openfasttrack/cli/TestCommandLineInterpreter.java
redcatbear: Removed since it does not matter whether the remaining parameters are detected correctly when an exception is thrown.

src/main/java/openfasttrack/cli/CommandLineInterpreter.java
redcatbear: More efficient but not much more readable.

src/main/java/openfasttrack/cli/CommandLineInterpreter.java
```diff
@@ -121,19 +121,25 @@ else if (argument.startsWith(SINGLE_CHAR_ARG_PREFIX))
...
-        final String arguments = argument.replaceFirst(SINGLE_CHAR_ARG_PREFIX, "").toLowerCase();
+        final String characters = argument.replaceFirst(SINGLE_CHAR_ARG_PREFIX, "").toLowerCase();
+        final int lastPosition = characters.length() - 1;
 
-        int position = 0;
-        for (final String character : arguments.split(""))
+        for (int position = 0; position <= lastPosition; ++position)
         {
-            ++position;
+            final String character = characters.substring(position, position + 1);
```
kaklakariada: I thought of using String.charAt(), but you want a String, so this is also fine :)

PR #47, opened 2017-01-21
0.1.0

redcatbear:
Release 0.1.0


.classpath
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kaklakariada: .classpath should be ignored and not checked in.

.project
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kaklakariada: .project should be ignored

.settings/org.eclipse.m2e.core.prefs
```diff
@@ -0,0 +1,4 @@
+activeProfiles=
```
kaklakariada: Not sure, but i think this should also be ignored

doc/.gitignore
```diff
@@ -0,0 +1,2 @@
+/system_requirements.html
```
kaklakariada: Move these lines to the 'global' .gitignore and remove this file

.classpath
kaklakariada: .classpath should be removed and ignored

kaklakariada:
Version 0.1.0 was already released (https://bintray.com/kaklakariada/maven/openfasttrack/0.1.0). Please use another version and update the version in pom.xml

Issue #53, opened 2017-01-24
TagImporter generates `covers` attribute


kaklakariada:
TagImporter shall create spec items with a generated ID and `covers` attribute set correctly.

See #49 for urther improvements.

PR #54, opened 2017-01-24
TagImporter generates covers attribute

kaklakariada:
See #53

### DIFF
--- a/src/main/java/openfasttrack/importer/tag/TagImporter.java
+++ b/src/main/java/openfasttrack/importer/tag/TagImporter.java
@@ -25,9 +25,11 @@ package openfasttrack.importer.tag;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.Reader;
+import java.nio.charset.StandardCharsets;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.zip.CRC32;
 
 import openfasttrack.core.SpecificationItemId;
 import openfasttrack.importer.ImportEventListener;
@@ -90,10 +92,10 @@ class TagImporter implements Importer
         while (matcher.find())
         {
             this.listener.beginSpecificationItem();
-            final String generatedName = this.fileName + ":" + lineNumber + "-" + counter;
+            final SpecificationItemId coveredId = SpecificationItemId.parseId(matcher.group(2));
+            final String generatedName = generateName(coveredId, lineNumber, counter);
             final SpecificationItemId generatedId = SpecificationItemId.createId(matcher.group(1),
                     generatedName, 0);
-            final SpecificationItemId coveredId = SpecificationItemId.parseId(matcher.group(2));
 
             LOG.finest(() -> "File " + this.fileName + ":" + lineNumber + ": found '" + generatedId
                     + "' covering id '" + coveredId + "'");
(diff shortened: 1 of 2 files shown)


src/main/java/openfasttrack/importer/tag/TagImporter.java
```diff
@@ -82,15 +86,21 @@ public void runImport()
...
     {
         final Matcher matcher = this.tagPattern.matcher(line);
+        int counter = 0;
         while (matcher.find())
         {
             this.listener.beginSpecificationItem();
-            final SpecificationItemId id = SpecificationItemId.parseId(matcher.group(1));
+            final String generatedName = this.fileName + ":" + lineNumber + "-" + counter;
+            final SpecificationItemId generatedId = SpecificationItemId.createId(matcher.group(1),
+                    generatedName, 0);
```
redcatbear: ":" and "-" are not allowed in IDs. I admit that we probably should allow "-". Used in URIs ":" is a character with a special meaning though.

Also file names can contain non-ASCII characters and white spaces which we would need to mangle if we want to create an ID from it.

Instead deduct the ID from the covered ID instead since we keep file and line in meta fields anyway:

Input: `[impl->dsn~tag_import.auto_id~1]`
Generated ID: `impl~tag_import.auto_id~1` covers `dsn~tag_import.auto_id~1`

This way you can tell from the ID what the implementation is supposed to do.

src/main/java/openfasttrack/importer/tag/TagImporter.java
```diff
@@ -82,15 +86,21 @@ public void runImport()
...
 
-            LOG.finest(
-                    () -> "File " + this.fileName + ":" + lineNumber + ": found id '" + id + "'");
-            this.listener.setId(id);
+            LOG.finest(() -> "File " + this.fileName + ":" + lineNumber + ": found '" + generatedId
+                    + "' covering id '" + coveredId + "'");
+            this.listener.setId(generatedId);
+            this.listener.addCoveredId(coveredId);
             this.listener.endSpecificationItem();
+            counter++;
```
redcatbear: I would find the Names nicer if you reset the counter with each line and do not attach the counter value if zero.

src/main/java/openfasttrack/importer/tag/TagImporter.java
kaklakariada: Now I generate the name using the CRC32 from filename, line number, counter and covered id.

src/main/java/openfasttrack/importer/tag/TagImporter.java
kaklakariada: The counter is now only used when calculating the CRC32, no need to make it readable.

src/test/java/openfasttrack/importer/tag/TestTagImporter.java
```diff
@@ -213,9 +245,24 @@ private static SpecificationItemId id(final String artifactType, final String na
...
+        return new SpecificationItem.Builder() //
+                .id(generatedId) //
+                .addCoveredId(coveredId) //
+                .build();
+    }
+
+    private static String generateName(final SpecificationItemId coveredId, final int lineNumber,
+            final int counter)
     {
-        return new SpecificationItem.Builder().id(id).build();
```
redcatbear: Code duplication in test?

Looks a little bit like: Verify 1 = 1 :stuck_out_tongue_winking_eye: 

Fixed values would be better as expectation. Since filename is a constant that should be no problem.

kaklakariada: You are right. But this means a lot of tedious work for each of the many tests and will break if we change anything in the generator.

redcatbear: True. Accepted.

PR #67, opened 2017-02-15
Feature/62 set exit status according to tacing result

redcatbear:
This is quite a large change for the feature. But I had to fix "the one" test case in order to get green again.

### DIFF
--- a/src/main/java/openfasttrack/cli/commands/Performable.java
+++ b/src/main/java/openfasttrack/cli/commands/Performable.java
@@ -1,0 +1,36 @@
+package openfasttrack.cli.commands;
+
+/*
+ * #%L
+ * OpenFastTrack
+ * %%
+ * Copyright (C) 2016 - 2017 hamstercommunity
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ * 
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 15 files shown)


redcatbear:
Still go one part of the test in TestCliStarter.java that I don't like:

Since the standard working directory for the tests is the project base directory, everything under "doc/" is used when no arguments are given. This leads to indeterministic test results since the documentation is part of tracing.

Would be good if we could reliably change the current working directory before the test.

src/test/java/openfasttrack/cli/TestCliStarter.java
```diff
@@ -52,144 +57,243 @@
...
     {
-        return new String(Files.readAllBytes(file), StandardCharsets.UTF_8);
+        final Path file = this.outputFile;
+        try
+        {
+            return new String(Files.readAllBytes(file), StandardCharsets.UTF_8);
+        }
+        catch (final IOException e)
+        {
+            e.printStackTrace();
```
kaklakariada: Rethrow exception

src/main/java/openfasttrack/cli/commands/Performable.java
```diff
@@ -0,0 +1,36 @@
...
+ * Interface for OpenFastTrack commands.
+ */
+public interface Performable
+{
+    /**
+     * Run an OpenFastTrack command
+     * 
+     * @return <code>true</code> if the operation performed was successful.
+     */
+    public boolean run();
```
kaklakariada: `public` is default, could be omitted

redcatbear: As we discussed I find it more expressive and consequent to add "public".

redcatbear:
Did not use isEmptyString() because the test is against a ByteStream. Using toString() would yield "null". Testing for Zero stream size seems cleaner to me.

kaklakariada:
@redcatbear please resolve the merge conflicts ;)

PR #68, opened 2017-02-25
58: Added integration test, CLI options and usage file.

redcatbear:
Task-Url: https://github.com/hamstercommunity/openfasttrack/issues/58

### DIFF
--- a/src/main/java/openfasttrack/cli/CliArguments.java
+++ b/src/main/java/openfasttrack/cli/CliArguments.java
@@ -69,6 +69,17 @@ public class CliArguments
     }
 
     /**
+     * Set the output file path
+     * 
+     * @param outputFile
+     *            output file path
+     */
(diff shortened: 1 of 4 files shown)


src/main/java/openfasttrack/cli/CliArguments.java
```diff
@@ -130,6 +130,17 @@ public void setOutputFormat(final String outputFormat)
     }
 
     /**
+     * Set the output format for an export or report
+     * 
+     * @param outputFormat
+     *            the output format
+     */
+    public void setO(final String outputFormat)
```
kaklakariada: Do we need tests for the short argument versions? My paranoia says yes ;)

redcatbear: I am on it.

Issue #79, opened 2017-12-18
Use sub-modules to structure project


kaklakariada:

kaklakariada:
Closing, won't fix, see #80

PR #80, opened 2017-12-18
Maven submodules

kaklakariada:
See issue #79


.gitignore
```diff
@@ -1,5 +1,6 @@
-/target/
+/**/.classpath
+/**/.project
```
redcatbear: This is not a good idea since it will clear the .project file with each branch.
And this file does not hurt anyway.

kaklakariada:
Sub-modules causes trouble with circle.ci builds (https://circleci.com/gh/hamstercommunity/openfasttrack/232) and it makes the build more complicated.

PR #88, opened 2018-01-19
Bug/86 collect list of covered specification item that do not exist

redcatbear:
@kaklakariada, @tomuben: This is the heart of OFT, so I would appreciate if both of you could give it a look.


src/main/java/openfasttrack/core/Linker.java
```diff
@@ -113,35 +125,56 @@ private void linkIgnoringRevision(final LinkedSpecificationItem item,
...
+    }
+
+    private LinkedSpecificationItem findOrCreateStaleItem(final SpecificationItemId id)
+    {
+        LinkedSpecificationItem staleItem = this.staleIndex.get(id);
+        if (staleItem == null)
+        {
+            staleItem = new LinkedSpecificationItem(new SpecificationItem.Builder().id(id).build());
+            this.staleIndex.put(id, staleItem);
+        }
```
kaklakariada: You could also use [Map.putIfAbsent()](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#putIfAbsent-K-V-)

redcatbear: In this case I always would have to build the stale Item (i.e. the value) on speculation.

PR #95, opened 2018-01-25
Update system_requirements.md

chku2469:


kaklakariada:
Build fails because travis-ci does not configure sonar correctly. This may be because this pull request comes from an external committer. I already added @chku2469 to @itsallcode/hamsters

Issue #101, opened 2018-01-31
Import status, tags and location from ".xml" / ".oreqm"


redcatbear:
# Situation
Status, tags and title are missing from both ".xml" / ".oreqm" import. Location is missing from ".oreqm".

# Acceptance Criteria
1. Import status, tags, title and location from ReqM2 files.

sascha15432:
obsolete due to #110

PR #110, opened 2018-02-17
Story/102 export status tags and location to spec object format

redcatbear:
@sascha15432 contains changes from #101 to so please review that one first.


sascha15432:
works as expected on Windows and Mac OS

PR #123, opened 2018-04-01
Support relative paths for legacy tag importer

kaklakariada:
This is required for gradle plugin configuration

### DIFF
--- a/src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterConfig.java
+++ b/src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterConfig.java
@@ -1,46 +1,41 @@
 package org.itsallcode.openfasttrace.importer.legacytag;
 
-/*-
- * #%L
- \* OpenFastTrace
- * %%
- * Copyright (C) 2016 - 2018 hamstercommunity
- * %%
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as
- * published by the Free Software Foundation, either version 3 of the
- * License, or (at your option) any later version.
- * 
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- * 
- * You should have received a copy of the GNU General Public
- * License along with this program.  If not, see
- * <http://www.gnu.org/licenses/gpl-3.0.html>.
- * #L%
- */
-
 import static java.util.Collections.emptyList;
 
 import java.nio.file.Path;
 import java.util.List;
+import java.util.Optional;
 
+/**
+ * Configuration for the {@link LegacyTagImporter}.
+ */
 public class LegacyTagImporterConfig
 {
     private final List<PathConfig> pathConfigs;
-    private final Path basePath;
+    private final Optional<Path> basePath;
 
-    public LegacyTagImporterConfig()
-    {
-        this(null, emptyList());
-    }
-
-    public LegacyTagImporterConfig(final Path basePath, final List<PathConfig> pathConfigs)
+    /**
+     * Create a new configuration object.
+     * 
+     * @param basePath
+     *            the root directory of the project to import. All paths will be
+     *            relative to this path.
+     * @param pathConfigs
+     *            a list of {@link PathConfig} objects.
+     */
+    public LegacyTagImporterConfig(final Optional<Path> basePath,
+            final List<PathConfig> pathConfigs)
     {
         this.basePath = basePath;
         this.pathConfigs = pathConfigs;
+    }
+
+    /**
+     * Creates a new, empty configuration.
+     */
+    public static LegacyTagImporterConfig empty()
+    {
+        return new LegacyTagImporterConfig(Optional.empty(), emptyList());
     }
 
     public List<PathConfig> getPathConfigs()
@@ -48,7 +43,7 @@ public class LegacyTagImporterConfig
         return this.pathConfigs;
     }
 
-    public Path getBasePath()
+    public Optional<Path> getBasePath()
     {
         return this.basePath;
     }
--- a/src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterConfig.java
+++ b/src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterConfig.java
@@ -24,24 +24,32 @@ package org.itsallcode.openfasttrace.importer.legacytag;
 
 import static java.util.Collections.emptyList;
 
+import java.nio.file.Path;
 import java.util.List;
 
 public class LegacyTagImporterConfig
 {
     private final List<PathConfig> pathConfigs;
+    private final Path basePath;
 
     public LegacyTagImporterConfig()
     {
-        this(emptyList());
+        this(null, emptyList());
     }
 
-    public LegacyTagImporterConfig(final List<PathConfig> pathConfigs)
+    public LegacyTagImporterConfig(final Path basePath, final List<PathConfig> pathConfigs)
     {
+        this.basePath = basePath;
         this.pathConfigs = pathConfigs;
     }
 
     public List<PathConfig> getPathConfigs()
     {
         return this.pathConfigs;
+    }
+
+    public Path getBasePath()
+    {
+        return this.basePath;
     }
 }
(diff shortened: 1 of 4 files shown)


src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterConfig.java
```diff
@@ -24,24 +24,32 @@
...
 
 public class LegacyTagImporterConfig
 {
     private final List<PathConfig> pathConfigs;
+    private final Path basePath;
 
     public LegacyTagImporterConfig()
     {
-        this(emptyList());
+        this(null, emptyList());
```
redcatbear: Using "null" this way always makes my danger sense tingle. How about an "Optional".
Alternatively a extra method to see if the path is set so that you can avoid the null checks in the calling code?

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterConfig.java
```diff
@@ -24,24 +24,32 @@
 
 import static java.util.Collections.emptyList;
 
+import java.nio.file.Path;
 import java.util.List;
 
 public class LegacyTagImporterConfig
```
redcatbear: Add Java doc to all public classes and methods please.

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
```diff
@@ -39,43 +40,61 @@
...
+    private final Supplier<LegacyTagImporterConfig> config;
 
     public LegacyTagImporterFactory()
     {
-        this(defaultConfig.getPathConfigs());
+        this(() -> defaultConfig);
     }
 
-    public LegacyTagImporterFactory(final List<PathConfig> pathConfigs)
+    public LegacyTagImporterFactory(final LegacyTagImporterConfig config)
```
redcatbear: Why support two types of constructors that are so similar in effect but use different paradigms?

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
```diff
@@ -39,43 +40,61 @@
...
 
     @Override
-    public boolean supportsFile(final Path file)
+    public boolean supportsFile(final Path path)
     {
-        return findConfig(file).isPresent();
+        return findConfig(path).isPresent();
     }
 
     private Optional<PathConfig> findConfig(final Path file)
```
redcatbear: You changed the parameter to "path" in the calling function but kept is as "file" here.
Suggest using "path" uniformly.

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
```diff
@@ -39,43 +40,61 @@
...
     {
-        return this.pathConfigs.stream().filter(config -> config.matches(file)).findFirst();
+        final Path relativePath = makeRelative(file);
+        return this.config.get().getPathConfigs().stream() //
+                .filter(config -> config.matches(relativePath)) //
+                .findFirst();
```
redcatbear: What happens if multiple match?

redcatbear: Wouldn't a hash map be more efficient here than iterating all path configs with each find?

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
```diff
@@ -39,43 +40,61 @@
...
+    }
+
+    private Path makeRelative(final Path path)
+    {
+        final Path basePath = this.config.get().getBasePath();
+        if (basePath == null)
```
redcatbear: Please add method to check if path is set instead of null check.

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
```diff
@@ -39,43 +40,61 @@
...
+        {
+            return path;
+        }
+        return basePath.relativize(path);
```
redcatbear: Does the match still work if the paths in the path config are provided absolute?

src/test/java/org/itsallcode/openfasttrace/importer/legacytag/TestLegacyTagImporterFactory.java
```diff
@@ -67,61 +63,89 @@ public void setup()
...
+    }
+
+    private LegacyTagImporterConfig config(final String basePath, final PathConfig... pathConfigs)
+    {
+        return new LegacyTagImporterConfig(basePath == null ? null : Paths.get(basePath),
+                asList(pathConfigs));
     }
 
-    private Importer createImporter(final List<PathConfig> pathConfigs, final Path path)
+    private LegacyTagImporterConfig config(final PathConfig... pathConfigs)
```
redcatbear: Please rename to "configure" to be in line with the Java naming standards.

Also to avoid mix-up of this method with variables of the same name.

src/test/java/org/itsallcode/openfasttrace/importer/legacytag/TestLegacyTagImporterFactory.java
```diff
@@ -67,61 +63,89 @@ public void setup()
...
+    }
+
+    private Importer createImporter(final LegacyTagImporterConfig config, final Path path)
+    {
+        return create(config).createImporter(path, StandardCharsets.UTF_8, this.listenerMock);
+    }
+
+    private LegacyTagImporterConfig config(final String basePath, final PathConfig... pathConfigs)
+    {
+        return new LegacyTagImporterConfig(basePath == null ? null : Paths.get(basePath),
```
redcatbear: The null values make this peace of code really hard to read. Especially without context.

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterConfig.java
kaklakariada: 👍 Using Optional.

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterConfig.java
kaklakariada: ✔️

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
kaklakariada: ✔️

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
kaklakariada: ✔️

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
kaklakariada: * For multiple matches we take the first matching path.
* We can't use a map because the matching logic is based on Java's path matcher.

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
kaklakariada: ✔️

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
kaklakariada: No. I created #127 to check for absolute paths.

src/test/java/org/itsallcode/openfasttrace/importer/legacytag/TestLegacyTagImporterFactory.java
kaklakariada: ✔️

src/main/java/org/itsallcode/openfasttrace/importer/legacytag/LegacyTagImporterFactory.java
redcatbear: Perfect. Thanks.

src/test/java/org/itsallcode/openfasttrace/importer/legacytag/TestLegacyTagImporterFactory.java
kaklakariada: ✔️

PR #151, opened 2018-06-23
Fix sonar and coverity warnings

kaklakariada:


src/main/java/org/itsallcode/openfasttrace/report/view/html/MarkdownLineStateMachine.java
```diff
@@ -28,6 +28,8 @@
 import java.util.List;
 import java.util.function.Function;
 
+// Duplicate strings help making this easier to understand.
+@SuppressWarnings("squid:S1192")
```
redcatbear: Please move suppression to `initializeTransitions()`

PR #154, opened 2018-07-01
Add available report formats to error message

kaklakariada:


src/main/java/org/itsallcode/openfasttrace/report/ReportFormat.java
```diff
@@ -55,6 +59,11 @@ public static ReportFormat parse(final String format)
                 return value;
             }
         }
-        throw new IllegalArgumentException("Unknown report format \"" + format + "\"");
+        final String availableValues = Arrays.stream(values()) //
```
redcatbear: Using a stream here is a bit much. Please use [Java 8 String.join()](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#join-java.lang.CharSequence-java.lang.Iterable-) instead.

kaklakariada: This only works for strings, not for enums. We need to keep this unless you want to change the input parameter from the short format (`plain`) to the enum value (`PLAIN_TEXT`)

Issue #163, opened 2018-07-17
Inline CSS into HTML report


redcatbear:
As a quality engineer
I want to deliver a tracing report in a single HTML file
so that the not so tech-savvy stakeholders are not bothered with keeping files together in order to read the report properly.

PR #164, opened 2018-07-21
feat(#163): Inline CSS stylesheet into HTML report

redcatbear:
Task-Url: https://github.com/itsallcode/openfasttrace/issues/163


src/main/java/org/itsallcode/openfasttrace/report/view/html/HtmlViewFactory.java
```diff
@@ -37,15 +38,17 @@
...
 
-    HtmlViewFactory(final PrintStream stream)
+    HtmlViewFactory(final PrintStream stream, final URL cssUrl)
     {
         super(stream);
+        this.cssUrl = cssUrl;
     }
 
-    public static HtmlViewFactory create(final OutputStream stream)
+    public static HtmlViewFactory create(final OutputStream stream, final URL cssURL)
```
kaklakariada: Inconsistent naming: `cssURL` vs `cssUrl`

src/test/java/org/itsallcode/openfasttrace/report/html/TestHtmlReportCssInlining.java
```diff
@@ -0,0 +1,80 @@
...
+    @Rule
+    public final SystemOutRule systemOutRule = new SystemOutRule().enableLog();
+
+    // [itest->dsn~reporting.html.inline_css~1]
+    @Test
+    public void testHtmlReportContainsInlineCSS() throws IOException
+    {
+        final File cssFile = this.tempFolder.newFile("test.css");
+        writeTextFile(cssFile, CSS);
+        final HtmlViewFactory factory = HtmlViewFactory.create(System.out, cssFile.toURI().toURL());
```
kaklakariada: In this case it is not necessary to use the workaround of SystemOutRule. Instead use a bytearrayinputstream from which you can directly get the data.

Issue #143, opened 2018-05-23
Refactor tag importers into a single class


kaklakariada:
**Is your feature request related to a problem? Please describe.**
When importing code tags from `*.java` files using the legacy tag importer the import fails because the 'normal' tag importer also uses `*.java`.

**Describe the solution you'd like**
A solution would be to allow configuring each importer for which files it is responsible.

**Describe alternatives you've considered**
Another option would be to order the importers by priority (giving the legacy tag importer higher priority) and use the first one without checking the others.

kaklakariada:
Implemented in #167

PR #167, opened 2018-08-02
Story/#143 unify short & long tag importer

kaklakariada:
See #143


src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -0,0 +1,91 @@
...
+import org.itsallcode.openfasttrace.importer.input.InputFile;
+
+// [impl->dsn~import.full-coverage-tag~1]
+class LongTagImportingLineConsumer implements LineConsumer
+{
+    private static final Logger LOG = Logger
+            .getLogger(LongTagImportingLineConsumer.class.getName());
+
+    private static final Pattern COVERED_ID_PATTERN = SpecificationItemId.ID_PATTERN;
+    private static final String COVERING_ARTIFACT_TYPE_PATTERN = "\\p{Alpha}+";
```
redcatbear: If it is a String please don't call it a "pattern", that should be reserved for pre-compiled patterns.

src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private final InputFile file;
+    private final ImportEventListener listener;
+    private final Pattern tagPattern;
+
+    LongTagImportingLineConsumer(final InputFile file, final ImportEventListener listener)
+    {
+        this.file = file;
+        this.listener = listener;
+        this.tagPattern = Pattern.compile(LONG_TAG_PATTERN_REGEX);
```
redcatbear: If I am not mistaken you could compile this as a static pattern already. It has no dynamic parts.

src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @Override
+    public void readLine(final int lineNumber, final String line)
+    {
+        final Matcher matcher = this.tagPattern.matcher(line);
+        int counter = 0;
+        while (matcher.find())
+        {
+            this.listener.beginSpecificationItem();
+            this.listener.setLocation(this.file.getPath(), lineNumber);
```
redcatbear: I think I just realized that longterm the column might be interesting too. But let's first see if we needed under real world conditions.

src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -0,0 +1,91 @@
...
+            final SpecificationItemId coveredId = SpecificationItemId.parseId(matcher.group(2));
```
redcatbear: You might consider to use names (i.e. constants) for the matching groups.
On the other hand the line is simple enough and has only one matching group, so it is not strictly necessary from a readability point of view.

src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -0,0 +1,91 @@
...
+            final String generatedName = generateName(coveredId, lineNumber, counter);
+            final SpecificationItemId generatedId = SpecificationItemId.createId(matcher.group(1),
+                    generatedName, 0);
+
+            LOG.finest(() -> "File " + this.file + ":" + lineNumber + ": found '" + generatedId
```
redcatbear: Since this is done quite often and has a lot of elements, a StringBuilder would be a good idea.
On the other hand it is log level FINEST.

src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -0,0 +1,91 @@
...
+            this.listener.addCoveredId(coveredId);
+            this.listener.endSpecificationItem();
+            counter++;
+        }
+    }
+
+    private String generateName(final SpecificationItemId coveredId, final int lineNumber,
+            final int counter)
+    {
+        final String uniqueName = this.file.getPath() + lineNumber + counter + coveredId.toString();
```
redcatbear: Use StringBuilder please. I know the situation got better with Java 8 and in theory the compiler should be able to internally use a StringBuilder for this line (no "+="), but better to be safe than sorry.

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporter.java
```diff
@@ -22,90 +22,54 @@
...
+            final ImportEventListener listener)
     {
-        this.file = file;
-        this.listener = listener;
-        this.tagPattern = Pattern.compile(TAG_PREFIX_PATTERN //
-                + "(" + COVERING_ARTIFACT_TYPE_PATTERN + ")" //
-                + "->" //
-                + "(" + COVERED_ID_PATTERN + ")" //
-                + TAG_SUFFIX_PATTERN);
+        this(createLineImporter(config, file, listener), file);
```
redcatbear: Hard to read. Please user intermediate variable "lineConsumer".

Also please use Consumer and Importer consistently.

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporter.java
```diff
@@ -22,90 +22,54 @@
...
-    private void processLine(final int lineNumber, final String line)
+    private static LineConsumer createLineImporter(final Optional<PathConfig> config,
+            final InputFile file, final ImportEventListener listener)
     {
-        final Matcher matcher = this.tagPattern.matcher(line);
-        int counter = 0;
-        while (matcher.find())
+        final List<LineConsumer> importers = new ArrayList<>();
+        importers.add(new LongTagImportingLineConsumer(file, listener));
+        if (config.isPresent())
```
redcatbear: :+1:

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporterFactory.java
```diff
@@ -21,24 +28,63 @@
...
  * {@link ImporterFactory} for tags in source code files.
  */
 // [impl->dsn~import.full-coverage-tag~1]
-public class TagImporterFactory extends RegexMatchingImporterFactory
+public class TagImporterFactory extends ImporterFactory
 {
-    public TagImporterFactory()
+    private static final Logger LOG = Logger.getLogger(TagImporterFactory.class.getName());
+
+    private static final String DEFAULT_FILE_REGEX = "(?i).*\\.java";
```
redcatbear: I think it would be a good idea to list as many popular programming language file extensions as a default here as possible.

Not all from that list but just as an inspiration:
https://www.file-extensions.org/filetype/extension/name/source-code-and-script-files

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporterFactory.java
```diff
@@ -21,24 +28,63 @@
...
+
+    public boolean supportsDefaultFile(final InputFile file)
+    {
+        return DEFAULT_FILE_PATTERN.matcher(file.getPath()).matches();
+    }
+
+    private Optional<PathConfig> findConfig(final InputFile file)
+    {
+        return getPathConfigs()//
+                .peek(c -> LOG.finest(() -> "Checking config " + c + " with file " + file))
```
redcatbear: Replace "c" with "config"

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporterFactory.java
```diff
@@ -21,24 +28,63 @@
...
     }
 
     @Override
-    public Importer createImporter(final InputFile file, final ImportEventListener listener)
+    public Importer createImporter(final InputFile path, final ImportEventListener listener)
+    {
+        if (!supportsFile(path))
+        {
+            throw new ImporterException(
+                    "File '" + path + "' not supported for import. Supported file name patterns: "
```
redcatbear: "' cannot be imported because it does not match any supported file patterns: "

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporterFactory.java
```diff
@@ -21,24 +28,63 @@
...
+                            + DEFAULT_FILE_REGEX + " and " + getPathConfigs().collect(toList()));
```
redcatbear: Actually you need a union here since there might be overlap. Question is if the user-defined config should not completely override the default (i.e. no merging).

src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
kaklakariada: done

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporterFactory.java
kaklakariada: Instead of hard coding a list we need to make this configurable. I created #168 for this.

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporterFactory.java
kaklakariada: Done.

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporterFactory.java
kaklakariada: Done.

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporterFactory.java
kaklakariada: This exception will never occur as no importer will be created if the file is not supported. This is just an additional assertion and the error message is not important.

src/main/java/org/itsallcode/openfasttrace/importer/tag/TagImporterFactory.java
redcatbear: Roger.

src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -0,0 +1,87 @@
...
+    private String generateName(final SpecificationItemId coveredId, final int lineNumber,
+            final int counter)
+    {
+        final String uniqueName = new StringBuilder() //
+                .append(this.file.getPath()) //
+                .append(lineNumber) //
+                .append(counter) //
+                .append(coveredId) //
+                .toString();
+        // this.file.getPath() + lineNumber + counter + coveredId.toString();
```
redcatbear: Remove commented out code.

PR #175, opened 2018-10-18
Fix self trace

kaklakariada:


.travis.yml
```diff
@@ -6,7 +6,6 @@ env:
 
 jdk:
   - oraclejdk8
-  - oraclejdk10
```
redcatbear: :+1:
Had the same problem in a different project recently. Thank you Oracle.

src/main/java/org/itsallcode/openfasttrace/core/SpecificationItemId.java
```diff
@@ -39,7 +39,7 @@
 {
     private static final Logger LOG = Logger.getLogger(SpecificationItemId.class.getName());
 
-    public static final String UNKONWN_ARTIFACT_TYPE = "unkonwn";
+    public static final String UNKNOWN_ARTIFACT_TYPE = "unknown";
```
redcatbear: Oh, that subtle. Good eyes.

src/test/java/org/itsallcode/openfasttrace/ITestSelfTrace.java
```diff
@@ -0,0 +1,71 @@
...
+    private Oft oft;
+
+    @Before
+    public void setUp()
+    {
+        this.oft = new OftRunner();
+    }
+
+    @Test
+    public void testSelfTrace()
```
redcatbear: :+1: * 100
That is a very elegant way to create a build breaker. I only came up with an additional build step but this is beautiful.

Issue #89, opened 2018-01-20
HTML Report


redcatbear:
As a developer
I want a tracing report that is easy to ready
so that I can identify and fix gaps and errors in the tracing easily.

As a quality engineer
I want a report that I can view in a browser
so that I don't have to install any special tools on my machine when auditing a project.

As a customer
I want a report that I can understand and archive
so that I get a solid view of whether the team working for me does a good job.

# Acceptance Criteria
* OFT can produce an HTML report from a trace
* The HTML report contains everything relevant for quality assurance

redcatbear:
Looking a template mechanisms:

Options

Engine | Last Commit | Repo | License | Runtime Dependencies | Notes
---------|------------------|--------|------------|-------------------------------|--------
Thymeleaf | 2017 | https://github.com/thymeleaf/thymeleaf | Apache 2.0 | javax.servlet, com.fasterxml.jackson.core, com.fasterxml.jackson.datatype |
Velocity | 2017 | http://svn.apache.org/viewvc/velocity/ | Apache 2.0 | slf4j, commons.lang |
MiniTemplator | 2012 | http://svn.source-code.biz/minitemplator-java/trunk/ | LGPL, EPL | none |
Jamon | 2012 | http://svn.code.sf.net/p/jamon/code/ | none | 2 stage, generates code from template

redcatbear:
Jdynamite looks promissing so far.

sascha15432:
+1 for Jdynamite but than we are going to violate the rule, that we don't pull in external dependencies at compile time. May we have to make a expection here for include a HTML template engine.

PR #201, opened 2018-12-01
Feature/89 trace summary in html report

redcatbear:
## Features
* #89: HTML tracing report now has summary and ToC.

### DIFF
--- a/src/main/java/org/itsallcode/openfasttrace/report/view/AbstractStreamableViewContainer.java
+++ b/src/main/java/org/itsallcode/openfasttrace/report/view/AbstractStreamableViewContainer.java
@@ -30,7 +30,7 @@ import java.io.PrintStream;
  */
 public abstract class AbstractStreamableViewContainer extends AbstractViewContainer
 {
-    protected PrintStream stream;
+    protected final PrintStream stream;
 
     public AbstractStreamableViewContainer(final PrintStream stream)
     {
--- a/src/main/java/org/itsallcode/openfasttrace/report/view/AbstractStreamableViewContainer.java
+++ b/src/main/java/org/itsallcode/openfasttrace/report/view/AbstractStreamableViewContainer.java
@@ -1,4 +1,4 @@
-package org.itsallcode.openfasttrace.report.view.html;
+package org.itsallcode.openfasttrace.report.view;
 
 /*-
  * #%L
@@ -22,20 +22,31 @@ package org.itsallcode.openfasttrace.report.view.html;
  * #L%
  */
 
-/**
- * Contains static helper methods for indentation.
- */
-public final class IndentationHelper
-{
-    private static final int INDENT_SPACES_PER_LEVEL = 2;
+import java.io.PrintStream;
 
-    private IndentationHelper()
+/**
+ * Abstract base class for View Containers that can be rendered to an output
+ * stream.
+ */
+public abstract class AbstractStreamableViewContainer extends AbstractViewContainer
+{
+    protected PrintStream stream;
+
+    public AbstractStreamableViewContainer(final PrintStream stream)
     {
-        // prevent instantiation.
+        super();
+        this.stream = stream;
     }
 
-    public static String createIndentationPrefix(final int level)
+    public AbstractStreamableViewContainer(final PrintStream stream, final String id,
+            final String title)
     {
-        return new String(new char[level * INDENT_SPACES_PER_LEVEL]).replace("\0", " ");
+        super(id, title);
+        this.stream = stream;
+    }
+
+    public void renderIndentation(final int level)
+    {
+        this.stream.print(IndentationHelper.createIndentationPrefix(level));
     }
 }
--- a/src/main/java/org/itsallcode/openfasttrace/report/view/AbstractViewContainer.java
+++ b/src/main/java/org/itsallcode/openfasttrace/report/view/AbstractViewContainer.java
@@ -47,7 +47,7 @@ public abstract class AbstractViewContainer implements ViewableContainer
      * 
      * @param id
      *            unique container ID
-     * @param title
+     * @param title container title
      */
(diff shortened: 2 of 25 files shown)


src/main/java/org/itsallcode/openfasttrace/report/view/AbstractStreamableViewContainer.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.io.PrintStream;
+
+/**
+ * Abstract base class for View Containers that can be rendered to an output
+ * stream.
+ */
+public abstract class AbstractStreamableViewContainer extends AbstractViewContainer
+{
+    protected PrintStream stream;
```
kaklakariada:
```suggestion
    protected final PrintStream stream;
```

src/main/java/org/itsallcode/openfasttrace/report/view/AbstractStreamableViewContainer.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public AbstractStreamableViewContainer(final PrintStream stream)
+    {
+        super();
```
kaklakariada:
```suggestion
```

src/main/java/org/itsallcode/openfasttrace/report/view/AbstractViewContainer.java
```diff
@@ -31,20 +31,49 @@
...
 
-    /*
-     * (non-Javadoc)
+    /**
+     * Create a new instance of type {@link AbstractViewContainer}.
      * 
-     * @see org.itsallcode.openfasttrace.report.view.Viewable#render(int)
+     * @param id
+     *            unique container ID
+     * @param title
```
kaklakariada:
```suggestion
     * @param title container title
```

src/main/java/org/itsallcode/openfasttrace/report/view/AbstractViewContainer.java
```diff
@@ -31,20 +31,49 @@
...
+    protected final String id;
+    protected final String title;
 
     /**
      * Create a new instance of type {@link AbstractViewContainer}.
      */
     public AbstractViewContainer()
     {
-        this.children = new ArrayList<>();
+        this(null, null);
```
kaklakariada: If id can be null: maybe make it Optional<String>?

src/main/java/org/itsallcode/openfasttrace/report/view/IndentationHelper.java
```diff
@@ -34,8 +34,15 @@ private IndentationHelper()
         // prevent instantiation.
     }
 
+    /**
+     * Create indentation prefix (i.e. white spaces)
+     * 
+     * @param level
+     *            indentation level
+     * @return white spaces
```
kaklakariada:
```suggestion
     * @return <code>level</code> white spaces
```

src/main/java/org/itsallcode/openfasttrace/report/view/html/HtmlReportSummary.java
```diff
@@ -0,0 +1,56 @@
...
+     * @param stream
+     *            print stream
+     */
+    public HtmlReportSummary(final PrintStream stream)
+    {
+        super(stream);
+    }
+
+    @Override
+    protected void renderBeforeChildren(final int level)
```
kaklakariada: Idea for refactoring: indentation is a cross cutting concern, maybe extract it into a base class. Maybe ViewContainer return HtmlElement objects with element name, attributes and children that are then rendered?

src/test/java/org/itsallcode/openfasttrace/report/view/html/TestHtmlReportDetails.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.itsallcode.openfasttrace.report.view.Viewable;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class TestHtmlReportDetails extends AbstractTestHtmlRenderer
+{
+    @Override
+    @BeforeEach
+    public void prepareEachTest()
```
kaklakariada: Is this necessary? The parent class method should be automatically called?

src/test/java/org/itsallcode/openfasttrace/report/view/html/TestHtmlTraceSummary.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private void renderTaceSummaryOnIndentationLevel(final int indentationLevel)
+    {
+        final Viewable view = this.factory.createTraceSummary(this.traceMock);
+        view.render(indentationLevel);
+    }
+
+    @ParameterizedTest
+    @ValueSource(ints = { 0, 1, 50, 99 })
```
kaklakariada: 👍

src/main/java/org/itsallcode/openfasttrace/report/view/AbstractViewContainer.java
redcatbear: If it was part of the external API, yes. But `null` does not appear on the outside, so I think this is okay.

src/test/java/org/itsallcode/openfasttrace/report/view/html/TestHtmlReportDetails.java
redcatbear: That was done by automatic refactoring. Good that you saw it, I completely missed that part.

src/main/java/org/itsallcode/openfasttrace/report/view/html/HtmlReportSummary.java
redcatbear: I think in the long run we will go for a template solution anyway. I would rather invest the time into that.

Issue #187, opened 2018-11-04
Migrate to Java 11


redcatbear:
As a developer
I want OFT to be built with Java
in order to avoid problems caused by the end of public updates for Java 8 in January 2019.

## Context
See https://www.oracle.com/technetwork/java/eol-135779.html for information about the EOL plans for the Oracle JDK.

Realistically with JDK 11 support for the Oracle JDK will not be guaranteed anymore by the OFT project due to Oracle's announcements to not provide it for free anymore. We recommend our users run OFT on the OpenJDK instead.

kaklakariada:
Migration to Java 11 already done.

PR #206, opened 2018-12-19
#187: Migrate to Java 11

kaklakariada:


api/src/main/java/org/itsallcode/openfasttrace/FilterSettings.java
```diff
@@ -24,208 +24,178 @@
...
-         * Build an instance of type {@link FilterSettings}.
-         * 
-         * @return the new instance.
-         */
-        public FilterSettings build()
-        {
-            return new FilterSettings(this);
-        }
-    }
+public final class FilterSettings {
```
redcatbear: Here something with the line ending formatter must have happended. This class should not be completely changed. Please check.

PR #207, opened 2019-01-12
Specification item import from PDF

tjaehnel:
This is a first attempt to import specification items from PDF.
It uses apache pdfbox to convert the PDF to text and then extracts specification items based on a regular expression.
Currently this is hardcoded to requirements in the form of [ABCD-123].

I want to make things like the pattern, artifact type, revision number, needs etc. configurable but I could not find out how to send parameters from the commandline to the importer.
If this is at all possible, could someone give me a hint, please?

Another thing I don't understand is why the Title does not appear in the converted output XML even though I'm calling setTitle().

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -116,6 +116,11 @@
             <version>3.0.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.pdfbox</groupId>
+            <artifactId>pdfbox</artifactId>
+            <version>2.0.13</version>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/src/main/java/org/itsallcode/openfasttrace/importer/pdf/PdfFileImporter.java
+++ b/src/main/java/org/itsallcode/openfasttrace/importer/pdf/PdfFileImporter.java
@@ -1,0 +1,107 @@
+package org.itsallcode.openfasttrace.importer.pdf;
+
+import java.io.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.pdfbox.pdmodel.PDDocument;
+import org.apache.pdfbox.text.PDFTextStripper;
+import org.itsallcode.openfasttrace.core.SpecificationItemId;
+/*-
+ * #%L
+ * OpenFastTrace
+ * %%
+ * Copyright (C) 2016 - 2018 itsallcode.org
+ * %%
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ * 
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public
+ * License along with this program.  If not, see
+ * <http://www.gnu.org/licenses/gpl-3.0.html>.
+ * #L%
+ */
+import org.itsallcode.openfasttrace.importer.*;
+import org.itsallcode.openfasttrace.importer.input.InputFile;
+
+/**
+ * This {@link Importer} supports reading PDF files
+ */
+public class PdfFileImporter implements Importer
+{
+    final static String DEFAULT_SPECITEM_PATTERN = "\\[(\\w+-\\d+)\\](.*)";
+    final static String DEFAULT_ARTIFACT_TYPE = "req";
+    final static int DEFAULT_SPECITEM_REVISION = 1;
+
+    private final InputFile file;
+    private final ImportEventListener listener;
+    protected final Pattern specItemPattern;
+
+    PdfFileImporter(final ImporterService importerService, final InputFile file,
+            final ImportEventListener listener)
+    {
+        this.file = file;
+        this.listener = listener;
+        this.specItemPattern = Pattern.compile(DEFAULT_SPECITEM_PATTERN);
+    }
+
+    @Override
+    public void runImport()
+    {
+        if (!this.file.isRealFile())
+        {
+            throw new UnsupportedOperationException(
+                    "Importing a PDF file from a stream is not supported");
+        }
+        try
+        {
+            final PDFTextStripper textStripper = new PDFTextStripper();
+            final PDDocument pdDoc = PDDocument.load(new File(this.file.getPath()));
+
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -116,6 +116,11 @@
             <version>3.0.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
redcatbear: Hi @tjaehnel, I'd like to move this importer to a separate repository, since we want to have no runtime dependencies except the Java SDK for the OpenFastTrace core. @kaklakariada created the service loader so that importers and exporters can be added at program start from regular JAR files. Once we switch to Java 11 (#206), this will be even more comfortable. Basically the importer JAR simply needs to be present in the class path at start-up.

src/main/java/org/itsallcode/openfasttrace/importer/pdf/PdfFileImporter.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+import org.itsallcode.openfasttrace.importer.*;
+import org.itsallcode.openfasttrace.importer.input.InputFile;
+
+/**
+ * This {@link Importer} supports reading PDF files
+ */
+public class PdfFileImporter implements Importer
+{
+    final static String DEFAULT_SPECITEM_PATTERN = "\\[(\\w+-\\d+)\\](.*)";
```
redcatbear: Using non-standard specification item ID notation makes manual debugging harder when you want to copy-paste IDs from a trace report. I therefore recommend to use standard IDs.

src/main/java/org/itsallcode/openfasttrace/importer/pdf/PdfFileImporter.java
```diff
@@ -0,0 +1,107 @@
...
+    {
+        if (!this.file.isRealFile())
+        {
+            throw new UnsupportedOperationException(
+                    "Importing a PDF file from a stream is not supported");
+        }
+        try
+        {
+            final PDFTextStripper textStripper = new PDFTextStripper();
+            final PDDocument pdDoc = PDDocument.load(new File(this.file.getPath()));
```
redcatbear: Please don't use abbreviations. Makes the code unnecessary hard to read.

src/main/java/org/itsallcode/openfasttrace/importer/pdf/PdfFileImporter.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Override
+    public void runImport()
+    {
+        if (!this.file.isRealFile())
+        {
+            throw new UnsupportedOperationException(
+                    "Importing a PDF file from a stream is not supported");
+        }
+        try
```
redcatbear: Make "else" explicit. Extract "else" body to method.

src/main/java/org/itsallcode/openfasttrace/importer/pdf/PdfFileImporter.java
```diff
@@ -0,0 +1,107 @@
...
+            throw new ImporterException("Error reading \"" + this.file + "\"", e);
+        }
+    }
+
+    protected void createSpecificationItemFromLine(final String line)
+    {
+        final Matcher matcher = this.specItemPattern.matcher(line);
+        if (matcher.find())
+        {
+            final String specItemName = matcher.group(1);
```
redcatbear: Use Constants for matcher group numbers.

src/main/java/org/itsallcode/openfasttrace/importer/pdf/PdfFileImporter.java
```diff
@@ -0,0 +1,107 @@
...
+            final String specItemTitle = matcher.group(2);
+
+            this.listener.beginSpecificationItem();
+
+            final SpecificationItemId specItemId = SpecificationItemId
+                    .createId(DEFAULT_ARTIFACT_TYPE, specItemName, DEFAULT_SPECITEM_REVISION);
+
+            this.listener.setId(specItemId);
+
+            this.listener.setTitle(specItemTitle.trim());
```
redcatbear: Exclude white spaces from regex pattern and remove trim().

redcatbear:
Summary:
1. Contains runtime dependency --> we will create a separate repository for this importer
1. Please document all public methods with JavaDoc.
1. Make methods more compact: Extract method where applicable and remove newlines.

redcatbear:
@tjaehnel: Created new repo: https://github.com/itsallcode/openfasttrace-pdf-importer
It's all your's!

redcatbear:
@kaklakariada is checking the build failures.

kaklakariada:
Sonar fails for external pull requests from forks. The only solution is to use internal pull requests (https://docs.travis-ci.com/user/sonarcloud/#analysis-of-internal-pull-requests).
@tjaehnel Please move the code to https://github.com/itsallcode/openfasttrace-pdf-importer, you @redcatbear gave you permissions for this repo.

src/main/java/org/itsallcode/openfasttrace/importer/pdf/PdfFileImporter.java
tjaehnel: Sorry I didn't get this comment.
What do you mean by "non-standard"? This is the Requirement ID format of one of the specifications I want to import. There are others as well which are completely different.
Also I understand from the user guide that specification item names can contain letters, numbers, hyphen, underscore and dot, which should be ensured with this pattern.

src/main/java/org/itsallcode/openfasttrace/importer/pdf/PdfFileImporter.java
tjaehnel: Very good point.

tjaehnel:
@kaklakariada thanks for that.
I'm holding off from moving the branch at the moment until my original questions are resolved and I know if what I'm planning to do is at all possible.

src/main/java/org/itsallcode/openfasttrace/importer/pdf/PdfFileImporter.java
redcatbear: After talking to @kaklakariada I see clearer what the reason for the PDF importer is: a spec you get as delivery from a 3rd party. In that case forget my comment since it is clear that you have to live with what the delivery contains.

pom.xml
tjaehnel: @redcatbear can you give me a hint how to register with the service loader and where to put the jar file until transition to Java 11 is done?

redcatbear: In your JAR archive add a property file called `org.itsallcode.openfasttrace.exporter.ExporterFactory` as resource. The content of all files of this name is automatically merged. The JAR itself needs to be in the class path at start time.

kaklakariada:
@tjaehnel any update on this? Should I close this?

Issue #224, opened 2019-02-01
Simplify Maven setup


kaklakariada:
Currently each sub-module must repeat the version number of the parent pom. Since Maven 3.5.0 this is not necessary any more, see section "Multi Module Setup" of https://maven.apache.org/maven-ci-friendly.html

kaklakariada:
Closed with #225

PR #225, opened 2019-02-03
#224 simplify maven setup

kaklakariada:


parent/pom.xml
```diff
@@ -4,11 +4,11 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.itsallcode</groupId>
     <artifactId>openfasttrace-parent</artifactId>
-    <version>2.2.0</version><!-- PRODUCT VERSION: do not remove this comment! -->
+    <version>${revision}</version>
     <name>OpenFastTrace Parent</name>
     <packaging>pom</packaging>
     <properties>
-        <oft.version>2.2.0</oft.version>
+        <revision>2.3.0-SNAPSHOT</revision>
```
redcatbear: Why is there a SNAPSHOT version here?

parent/pom.xml
```diff
@@ -429,6 +472,25 @@
                                         <ignore></ignore>
                                     </action>
                                 </pluginExecution>
+                                <pluginExecution>
+                                    <pluginExecutionFilter>
+                                        <groupId>
```
redcatbear: Please correct (unify) XML formatting.

redcatbear: `<groupId>org.codehaus.mojo</groupId>` and so on.

kaklakariada: OK. Can you please provide the XML formatter settings so that we can document this?

PR #232, opened 2019-03-31
Bugfix/231 fix maven deployment

kaklakariada:


api/pom.xml
```diff
@@ -6,19 +6,14 @@
     <name>OpenFastTrace API</name>
     <parent>
         <relativePath>../parent/pom.xml</relativePath>
-        <groupId>org.itsallcode</groupId>
+        <groupId>org.itsallcode.openfasttrace</groupId>
```
redcatbear: Okay. Works for me. Looks a little bit strange at first glance, but makes sense.

PR #261, opened 2021-04-26
260 Aspec Reporter

poldi2015:
The Aspec Report is a new reporter with name "apec" that reports all items and their coverage and linkage as XML file so that that tracing renerated by openfasttrace can be processed by other tools.

Closes #260

PR #258, opened 2021-04-16
Handling of files for which no importer (or too many) is present does not fail with an exception

poldi2015:
- [x] Implement files for ZIP File Importer
- [x] Improve ZIP File importer
- [x] Implement tests

PR #272, opened 2021-05-24
Prepare release 3.2.0

kaklakariada:
This prepares for the 3.2.0 release and removes some unused files.
The next release includes the following PR:
* #261 
* #258

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,4 +1,0 @@
-coverage:
-  range: 70..95
-  round: down
-  precision: 2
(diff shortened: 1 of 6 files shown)


.codecov.yml
```diff
@@ -1,4 +0,0 @@
-coverage:
```
kaklakariada, line 1: Not used any more, we replaced codecove with sonar.

Jenkinsfile
```diff
@@ -1,7 +0,0 @@
-node {
```
kaklakariada, line 1: This was only added for testing. CI is now done with Github Actions.

morazow (approved):
:+1:

PR #273, opened 2021-05-24
Speed up build

kaklakariada:
Run parallel maven build, using one thread per cpu core

See https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3


kaklakariada:
No improvement of Github actions build speed.

PR #295, opened 2021-08-27
Using Temurin builds of OpenJDK in GH actions

carldea:
The AdoptOpenJDK has been discontinued Since July 2021. When using Zulu you get all the latest updated builds (TCK tested) for all versions of OpenJDK. 🥇

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
 
     - uses: actions/setup-java@v2
       with:
-        distribution: 'zulu'
+        distribution: 'temurin'
         java-version: ${{ matrix.java }}
 
     - name: Cache local Maven repository
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
 
     - uses: actions/setup-java@v2
       with:
-        distribution: 'adopt'
+        distribution: 'zulu'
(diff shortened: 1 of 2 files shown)


kaklakariada:
Thank you for the pr!

See https://github.com/marketplace/actions/setup-java-jdk#supported-distributions for details.

@redcatbear which jdk should we use: Zulu or Eclipse temurin?

kaklakariada:
We have decided to use Eclipse Temurin, because it it is a community-driven Project and the successor of AdoptOpenJdk.
@carldea : please update the PR to use `temurin`. Thank you!

.github/workflows/build.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
     - uses: actions/setup-java@v2
       with:
-        distribution: 'adopt'
+        distribution: 'zulu'
```
kaklakariada, line 28:
```suggestion
        distribution: 'temurin'
```
See comment. We prefer to use Eclipse Temurin

kaklakariada (changes requested):

.github/workflows/build.yml
carldea: I've updated both workflows to use Temurin as the distribution for setup-java GH actions. 👍🏼

kaklakariada (approved):

PR #298, opened 2021-08-31
Fix sonar issues

kaklakariada:


reporter/html/src/test/java/org/itsallcode/openfasttrace/report/html/view/html/MarkdownSpanConverterTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+class MarkdownSpanConverterTest
+{
+    @ParameterizedTest(name = "Line ''{0}'' converted to HTML ''{1}''")
+    @CsvSource(
+    {
+            "'    code\n', '<pre>    code\n</pre>'",
+            "'    code', '    code'",
```
kaklakariada, line 37: The implementation seems to be broken here. I would expect output `<pre>    code</pre>` in this case.

@redcatbear Please create an issue if you think this is a problem.

redcatbear: Actually it should be n-4 spaces in the `pre` block.

reporter/html/src/test/java/org/itsallcode/openfasttrace/report/html/view/html/OriginLinkFormatterTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.itsallcode.openfasttrace.api.core.Location;
+import org.junit.jupiter.api.Test;
+
+class OriginLinkFormatterTest
+{
+    @Test
+    void formatNormalPath()
+    {
+        assertFormatter(Location.create("path", 4), "path:4");
```
redcatbear, line 36: Wrap in `assertAll`, please.

reporter/html/src/test/java/org/itsallcode/openfasttrace/report/html/view/html/OriginLinkFormatterTest.java
```diff
@@ -0,0 +1,70 @@
...
+    void formatUrls()
+    {
+        assertLinkFormatter(Location.create("http://server:80/path", 4));
+        assertLinkFormatter(Location.create("https://server:443/path", 4));
+        assertLinkFormatter(Location.create("ftp://server:21/path", 4));
+        assertLinkFormatter(Location.create("file://server/path", 4));
+        assertLinkFormatter(Location.create("mailto:user@example.com", 4));
+    }
+
+    private void assertLinkFormatter(Location location)
```
redcatbear, line 53: Having fixed strings would be better here even if that means code duplication. Otherwise you test an algorithm against itself.

redcatbear (changes requested):

reporter/html/src/test/java/org/itsallcode/openfasttrace/report/html/view/html/MarkdownSpanConverterTest.java
kaklakariada: This is more effort to fix. I created ticket https://github.com/itsallcode/openfasttrace/issues/312

reporter/html/src/main/java/org/itsallcode/openfasttrace/report/html/view/html/OriginLinkFormatter.java
```diff
@@ -138,6 +138,6 @@ private static URI convertPathWithoutProtocolToUri(final String path)
 
     private static String unWindowsify(final String path)
     {
-        return path.replaceAll("\\\\", "/");
+        return path.replace("\\\\", "/");
```
redcatbear, line 141: :+1: Good one. That is definitely more efficient for a literal. I think I instinctively thought that `replace` would only search for one occurrence like in grep without modifiers. Helps to read the API, I guess.

redcatbear (approved):

PR #386, opened 2024-02-16
#378: RST importer

### LINKED ISSUE #378: Support for RST importer
**Description**

These days more and more documents are written in ReST format (especially in the open source projects), whereas at the moment, OFT expects a document to be written in Markdown format. It would be great to meet users' expectations by adding support for the new format which is more powerful and supports more complex data structures. Despite many similarities, converting from one format to another takes a long time.

**Intermediate solution**

With the following modification, we can hack OFT to detect RST files as if they were MD files. However, due to differences in heading format, the tool is no longer able to detect the title.
```
diff --git a/importer/markdown/src/main/java/org/itsallcode/openfasttrace/importer/markdown/MarkdownImporterFactory.java b/importer/markdown/src/main/java/org/itsallcode/openfasttrace/importer/markdown/MarkdownImporterFactory.java
index d112694afba5..f1c068fe2a85 100644
--- a/importer/markdown/src/main/java/org/itsallcode/openfasttrace/importer/markdown/MarkdownImporterFactory.java
+++ b/importer/markdown/src/main/java/org/itsallcode/openfasttrace/importer/markdown/MarkdownImporterFactory.java
@@ -11,7 +11,7 @@ public class MarkdownImporterFactory extends RegexMatchingImporterFactory
     /** Creates a new instance. */
     public MarkdownImporterFactory()
     {
-        super("(?i).*\\.markdown", "(?i).*\\.md");
+        super("(?i).*\\.markdown", "(?i).*\\.md", "(?i).*\\.rst");
     }

     @Override
```

redcatbear:
Closes #378.


parent/pom.xml
redcatbear, line 1: @kaklakariada, are you sure these changes don't break overall coverage reporting in the CI?

kaklakariada: I was pretty sure, but reverted the changes nevertheless 😅

importer/markdown/src/test/java/org/itsallcode/openfasttrace/importer/markdown/ITMarkdownMarkupImporter.java
```diff
@@ -0,0 +1,139 @@
...
+import org.hamcrest.Matcher;
+import org.itsallcode.openfasttrace.api.core.SpecificationItem;
+import org.itsallcode.openfasttrace.api.core.SpecificationItemId;
+import org.itsallcode.openfasttrace.api.importer.ImporterFactory;
+import org.itsallcode.openfasttrace.importer.lightweightmarkup.AbstractLightWeightMarkupImporterTest;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class ITMarkdownMarkupImporter extends AbstractLightWeightMarkupImporterTest {
```
kaklakariada, line 16: Could we rename this to `TestMarkdownMarkupImporter` so that it runs with normal unit tests? It should be quick enough as it doesn't touch files or network.

importer/markdown/src/test/java/org/itsallcode/openfasttrace/importer/markdown/ITMarkdownMarkupImporter.java
```diff
@@ -0,0 +1,139 @@
...
+                                 1))
+                         .title("This is a title with an underline")
+                         .description("Body text.")
+                         .location("y", 4)
+                         .build()));
+     }
+
+     @Test
+     void
+     testLessThenTwoUnderliningCharactersAreNotDetectedAsTitleUnderlines()
```
kaklakariada, line 127:
```suggestion
     testLessThanThreeUnderliningCharactersAreNotDetectedAsTitleUnderlines()
```

kaklakariada (changes requested):

PR #387, opened 2024-02-16
#222: Fix compiler warnings

### LINKED ISSUE #222: Fix maven warnings
Maven generates some warning messages during the build. Fix them.
Also fail compilation on warnings: `<arg>-Xlint:all</arg>` and `<arg>-Werror</arg>`

```xml
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <version>3.12.1</version>
                <configuration>
                    <source>${java.version}</source>
                    <target>${java.version}</target>
                    <showWarnings>true</showWarnings>
                    <compilerArgs>
                        <!-- We turn off `processing` because it complains about every annotation -->
                        <arg>-Xlint:all</arg>
                        <arg>-Werror</arg>
                    </compilerArgs>
                </configuration>
            </plugin>
```

see also https://github.com/exasol/project-keeper/blob/main/maven-project-crawler/pk_generated_parent.pom#L88

kaklakariada:
Closes #222


exporter/markdown/pom.xml
redcatbear, line 1: So we actually don't have a Markdown exporter. Ooops. I was wandering where that came from and what it would be good for...

redcatbear (approved):

PR #389, opened 2024-02-19
#379: Allow specifying revision for tags

### LINKED ISSUE #379: Support specifying a revision in code tags
**Description**

At the moment, it is not possible to specify a revision of a tag present in a code. The tool automatically assigns revision 0 in that case
and the name is inherited from the parent.
Example:
`[impl->req~example_name~1>>test]`
which results in:
`impl~example_name~0`

**Example of a new format**
`[impl~~1->req~example_name~1>>test]`
which results in:
`impl~example_name~1`

kaklakariada:
Closes #379

### DIFF
--- a/importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
+++ b/importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
@@ -21,16 +21,15 @@ class LongTagImportingLineConsumer extends RegexLineConsumer
 
     private static final String COVERING_ARTIFACT_TYPE_PATTERN = "\\p{Alpha}+";
     // [impl->dsn~import.full-coverage-tag-with-revision~1]
-    private static final String COVERING_TAG = "(" + COVERING_ARTIFACT_TYPE_PATTERN + ")"
-            + optional(SpecificationItemId.ARTIFACT_TYPE_SEPARATOR + SpecificationItemId.REVISION_SEPARATOR
-                    + SpecificationItemId.ITEM_REVISION_PATTERN);
     private static final String OPTIONAL_WHITESPACE = "\\s*";
     private static final String TAG_PREFIX = "\\[";
     private static final String TAG_SUFFIX = "\\]";
     private static final String NEEDS_COVERAGE = ">>" + OPTIONAL_WHITESPACE + "(\\p{Alpha}+(?:" + OPTIONAL_WHITESPACE
             + "," + OPTIONAL_WHITESPACE + "\\p{Alpha}+)*)";
     private static final String TAG_REGEX = TAG_PREFIX + OPTIONAL_WHITESPACE//
-            + COVERING_TAG //
+            + "(" + COVERING_ARTIFACT_TYPE_PATTERN + ")"
+            + optional(SpecificationItemId.ARTIFACT_TYPE_SEPARATOR + SpecificationItemId.REVISION_SEPARATOR
+                    + SpecificationItemId.ITEM_REVISION_PATTERN) //
             + OPTIONAL_WHITESPACE + "->" + OPTIONAL_WHITESPACE //
             + "(" + SpecificationItemId.ID_PATTERN + ")" //
             + OPTIONAL_WHITESPACE + optional(NEEDS_COVERAGE + OPTIONAL_WHITESPACE) //
@@ -59,7 +58,7 @@ class LongTagImportingLineConsumer extends RegexLineConsumer
         final SpecificationItemId coveredId = SpecificationItemId.parseId(matcher.group(3));
         final List<String> neededArtifactTypes = parseCommaSeparatedList(matcher.group(7));
         final String artifactType = matcher.group(1);
-        final Optional<String> revision = Optional.ofNullable(matcher.group(2));
+        final String revision = matcher.group(2);
         final SpecificationItemId generatedId = generateItemId(lineNumber, lineMatchCount, coveredId, artifactType,
                 revision, neededArtifactTypes);
 
@@ -93,12 +92,18 @@ class LongTagImportingLineConsumer extends RegexLineConsumer
     }
 
     private SpecificationItemId generateItemId(final int lineNumber, final int lineMatchCount,
-            final SpecificationItemId coveredId, final String artifactType, final Optional<String> revision,
+            final SpecificationItemId coveredId, final String artifactType, final String revision,
             final List<String> neededArtifactTypes)
     {
         final String name = getItemName(lineNumber, lineMatchCount, coveredId, neededArtifactTypes);
-        final int parsedRevision = revision.map(Integer::parseInt).orElse(0);
-        return SpecificationItemId.createId(artifactType, name, parsedRevision);
+        return SpecificationItemId.createId(artifactType, name, parseRevision(revision));
+    }
+
+    private int parseRevision(final String revision)
+    {
+        return Optional.ofNullable(revision)
+                .map(Integer::parseInt)
+                .orElse(0);
     }
 
     // [impl->dsn~import.full-coverage-tag-with-needed-coverage-readable-names~1]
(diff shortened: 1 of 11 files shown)


importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -21,13 +20,17 @@ class LongTagImportingLineConsumer extends RegexLineConsumer
...
+            + optional(SpecificationItemId.ARTIFACT_TYPE_SEPARATOR + SpecificationItemId.REVISION_SEPARATOR
+                    + SpecificationItemId.ITEM_REVISION_PATTERN);
     private static final String OPTIONAL_WHITESPACE = "\\s*";
     private static final String TAG_PREFIX = "\\[";
     private static final String TAG_SUFFIX = "\\]";
     private static final String NEEDS_COVERAGE = ">>" + OPTIONAL_WHITESPACE + "(\\p{Alpha}+(?:" + OPTIONAL_WHITESPACE
             + "," + OPTIONAL_WHITESPACE + "\\p{Alpha}+)*)";
     private static final String TAG_REGEX = TAG_PREFIX + OPTIONAL_WHITESPACE//
-            + "(" + COVERING_ARTIFACT_TYPE_PATTERN + ")" //
+            + COVERING_TAG //
```
redcatbear, line 33: This is a deviation from the scheme of the other regex elements. In all others we make the grouping brackets transparent (intended) here we include it into the covering tag fragement. I think you should revert this.

importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -53,10 +56,12 @@ public void processMatch(final Matcher matcher, final int lineNumber, final int
...
         this.listener.setLocation(this.file.getPath(), lineNumber);
-        final SpecificationItemId coveredId = SpecificationItemId.parseId(matcher.group(2));
-        final List<String> neededArtifactTypes = parseCommaSeparatedList(matcher.group(6));
-        final SpecificationItemId generatedId = generateItemId(lineNumber, lineMatchCount, coveredId, matcher.group(1),
-                neededArtifactTypes);
+        final SpecificationItemId coveredId = SpecificationItemId.parseId(matcher.group(3));
+        final List<String> neededArtifactTypes = parseCommaSeparatedList(matcher.group(7));
+        final String artifactType = matcher.group(1);
+        final Optional<String> revision = Optional.ofNullable(matcher.group(2));
+        final SpecificationItemId generatedId = generateItemId(lineNumber, lineMatchCount, coveredId, artifactType,
```
redcatbear, line 63: Passing `Optional` as method parameter considered harmful:
https://www.baeldung.com/java-optional#misuages

Money quote:
> The intent of Java when releasing Optional was to use it as a return type

There is a reason why I am not a fan o Optional. It does not remove the potential for the parameter being null for example. In essence, Optional should neither be used as parameter nor as class attribute. 

A cleaner way would be overloading methods. Even a `null` check is in the end more elegant.

importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/RegexLineConsumer.java
```diff
@@ -36,10 +36,11 @@ public void readLine(final int lineNumber, final String line)
      * @param matcher
      *            regular expression {@link Matcher}.
      * @param lineNumber
-     *            line number of the matched the input, starting with  {@code 1} for the first line.
+     *            line number of the matched the input, starting with {@code 1}
+     *            for the first line.
      * @param lineMatchCount
-     *            number of the current match in the context of the current line,
-     *            starting with {@code 0} for the first match in a line.
+     *            number of the current match in the context of the current
```
redcatbear, line 42: Strange. Line width for comments and code should be 100 chars. Why it this wrapping here?

importer/tag/src/test/java/org/itsallcode/openfasttrace/importer/tag/TestTagImporter.java
```diff
@@ -268,6 +268,29 @@ void tagWithMultipleRequiredCoverageWithSpaces()
                 itemWithReadableName(COVERING_ARTIFACT_TYPE1, 1, ID1, List.of(NEEDED_COVERAGE1, NEEDED_COVERAGE2)));
     }
 
+    // [utest->dsn~import.full-coverage-tag-with-revision~1]
+    @Test
+    void tagWithRevision()
+    {
+        assertItems(
+                tag("impl~~42", "req~example_name~17"), //
```
redcatbear, line 276:
```suggestion
                tag("impl~~42", "req~example_name~17"), //
```

While this is compact, it hides the actual test. I recommend writing out the complete tags text. This way the setup is much more explicit. Also, It is about the same number of characters to type. :-)

importer/tag/src/test/java/org/itsallcode/openfasttrace/importer/tag/TestTagImporter.java
```diff
@@ -268,6 +268,29 @@ void tagWithMultipleRequiredCoverageWithSpaces()
...
+                        .addCoveredId("req", "example_name", 17)
+                        .location(FILENAME, 1).build());
+    }
+
+    // [utest->dsn~import.full-coverage-tag-with-revision~1]
+    @Test
+    void tagWithRevisionAndNeededCoverage()
+    {
+        assertItems(
+                tag("impl~~42", "req~example_name~17", "test"), //
```
redcatbear, line 287:
```suggestion
                tag("impl~~42", "req~example_name~17", "test"), //
```
Here you see what I mean. This test is hard to decipher. If I didn't know about the forwarding feature, I would have zero clue what is tested here. Again, please spell the test out. You might even consider to provide the input as `ValueSource`, so that you can test a couple permutations for improved robustness.

product/src/test/java/org/itsallcode/openfasttrace/ITestSelfTrace.java
```diff
@@ -63,14 +59,19 @@ protected String createSelfTraceReport(final Trace trace)
...
         return ImportSettings.builder() //
                 .addInputs(baseDir.resolve("doc/spec/")) //
                 .addInputs(findInputDirectories(baseDir)) //
                 .build();
     }
 
-    private List<Path> findInputDirectories(Path rootProjectDir)
+    private Path getProjectBaseDir()
+    {
+        return Paths.get(new StandardDirectoryService().getCurrent()).resolve("..").normalize();
```
redcatbear, line 71: :+1: Yes, that is better!

redcatbear (changes requested):

importer/tag/src/test/java/org/itsallcode/openfasttrace/importer/tag/TestTagImporter.java
kaklakariada: I had this before, but it caused problems when tracing OFT's own requirements. The tag for testing was recognized as a real tag.

importer/tag/src/test/java/org/itsallcode/openfasttrace/importer/tag/TestTagImporter.java
```diff
@@ -272,8 +272,8 @@ void tagWithMultipleRequiredCoverageWithSpaces()
     @Test
     void tagWithRevision()
     {
-        assertItems(
-                tag("impl~~42", "req~example_name~17"), //
+        // Concatenation required to avoid problems during self-trace
+        assertItems("[impl~~42->req~example_name~17" + "]",
```
redcatbear, line 276:
```suggestion
        assertItems("[impl~~42->req~example_name~17" + "]",
```
:+1: I though about this last night. And that is exactly the solution that I had in mind at the end too. Dead simple and does the trick.

redcatbear (approved):
Thanks for the fixes. Well done!

PR #390, opened 2024-02-19
Move XML parser to separate module

kaklakariada:
## Situation

The XML parser is useful for multiple scenarios in which XML must be imported.

In this is ticket, we extract the XML parser to make it reusable.


importer/xmlparser/src/test/java/org/itsallcode/openfasttrace/importer/xmlparser/event/AttributeTest.java
```diff
@@ -0,0 +1,58 @@
...
+        assertThat(map.size(), is(0));
+    }
+
+    @Test
+    void buildMapWithSingleElement() {
+        when(attributesMock.getLength()).thenReturn(1);
+        when(attributesMock.getQName(0)).thenReturn("qname0");
+        when(attributesMock.getValue(0)).thenReturn("value0");
+        final Map<String, Attribute> map = Attribute.buildMap(attributesMock);
+        assertAll(() -> assertThat(map.size(), is(1)),
```
redcatbear, line 41: Alternatively:

`assertThat(map, contains(allOf(...equalTo...))`

importer/xmlparser/src/test/java/org/itsallcode/openfasttrace/importer/xmlparser/event/EndElementEventTest.java
```diff
@@ -0,0 +1,41 @@
...
+                () -> assertThat(qName.getPrefix(), equalTo("")));
+    }
+
+    @Test
+    void getLocation() {
+        assertThat(testee().getLocation(), sameInstance(LOCATION));
+    }
+
+    private EndElementEvent testee() {
+        return EndElementEvent.create(URI, QNAME, QNAME, LOCATION);
```
redcatbear, line 39: I am almost certain that the code would be more readable if you hadn't extracted this and used string literals instead of constants. Also, you could use different literals to spice things up. :-)

importer/xmlparser/src/test/java/org/itsallcode/openfasttrace/importer/xmlparser/tree/CallbackContentHandlerTest.java
```diff
@@ -0,0 +1,56 @@
...
+import org.xml.sax.Attributes;
+
+@ExtendWith(MockitoExtension.class)
+class CallbackContentHandlerTest {
+    private static final String ELEMENT_URI = "";
+    private static final String ELEMENT_LOCAL_NAME = "localName";
+    private static final String ELEMENT_QNAME = "qname";
+
+    @Mock
+    Consumer<TreeElement> listenerMock;
```
redcatbear, line 24: I would make this a parameter.

importer/xmlparser/src/test/java/org/itsallcode/openfasttrace/importer/xmlparser/tree/TreeElementTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Test
+    void isRootFalsForNonNullParent() {
+        final TreeElement element = new TreeElement(createStartElement(), mock(TreeElement.class));
+        assertThat(element.isRootElement(), is(false));
+    }
+
+    @Test
+    void getAttributeValueByName() {
+        assertThat(testee().getAttributeValueByName("qname0").getValue(), equalTo("value0"));
```
redcatbear, line 76: This is a typical example where the fact that the test preparation is extracted, makes the test harder to validate.

importer/xmlparser/src/test/java/org/itsallcode/openfasttrace/importer/xmlparser/tree/TreeElementTest.java
```diff
@@ -0,0 +1,121 @@
...
+    void invokeEndElementListenersSuccedsForEmptyList() {
+        assertDoesNotThrow(testee()::invokeEndElementListeners);
+    }
+
+    @Test
+    void invokeEndElementListenersCallsSingleEntry() {
+        final TreeElement element = testee();
+        element.addEndElementListener(listenerMock1);
+        element.invokeEndElementListeners();
+        verify(listenerMock1).accept(same(element));
```
redcatbear, line 94:
```suggestion
        verify(listenerMock1).accept(same(element));
```
Isn't the fact that it is the same instance testing for an implementation detail? Hm. Element is probably supposed to be mutable in most cases, so it might be okay as a test.

redcatbear (changes requested):

importer/xmlparser/src/main/java/org/itsallcode/openfasttrace/importer/xmlparser/ContentHandlerAdapter.java
```diff
@@ -44,27 +41,27 @@ public ContentHandlerAdapter(final String filePath, final XMLReader xmlReader,
...
+    void registerListener() {
+        verifyNoContentHandlerRegistered();
         this.delegate.init(this);
         this.xmlReader.setContentHandler(this);
     }
 
+    private void verifyNoContentHandlerRegistered() {
+        final ContentHandler originalContentHandler = this.xmlReader.getContentHandler();
+        if (originalContentHandler != null) {
+            throw new IllegalStateException("An XML content handler is already registered: " + originalContentHandler);
```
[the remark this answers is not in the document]
kaklakariada: Correct. But I prefer to fail early. Should I rather throw an `AssertionError`?

importer/xmlparser/src/test/java/org/itsallcode/openfasttrace/importer/xmlparser/tree/TreeElementTest.java
kaklakariada: Correct, this is testing an implementation detail. I am willing to accept the risk of having to change this test should the implementation change.
The `element` object is immutable.

importer/xmlparser/src/test/java/org/itsallcode/openfasttrace/importer/xmlparser/event/AttributeTest.java
kaklakariada: Sorry, I could not get this working

redcatbear (approved):

PR #395, opened 2024-02-26
#394: Import tags also from `.yml` files

### LINKED ISSUE #394: Import tags also from `.yml` files
OFT already imports tags from `.yaml` files (see class `TagImporterFactory`). Often YAML files use the shorter `.yml` suffix.

That's why OFT should also import tags from `.yml` files.

kaklakariada:
Closes #394


redcatbear:
I think this might be mentioned in the user guide too.

redcatbear (approved):

PR #403, opened 2024-03-10
#399: Configure and verify reproducible build

### LINKED ISSUE #399: Replace reproducible-build-maven-plugin with Maven configuration
Currently we use https://zlika.github.io/reproducible-build-maven-plugin/. But new Maven versions support reproducible builds without additional dependencies, see documentation at https://maven.apache.org/guides/mini/guide-reproducible-builds.html

To avoid unnecessary dependencies we should replace the plugin with correct configuration.

kaklakariada:
Closes #399

### DIFF
--- a/doc/developer_guide.md
+++ b/doc/developer_guide.md
@@ -178,6 +178,22 @@ mvn clean org.jacoco:jacoco-maven-plugin:prepare-agent package sonar:sonar \
 
 See analysis results at [sonarcloud.io](https://sonarcloud.io/dashboard?id=org.itsallcode.openfasttrace%3Aopenfasttrace).
 
+## Reproducible Build
+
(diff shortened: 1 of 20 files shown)


doc/developer_guide.md
```diff
@@ -178,6 +178,22 @@ mvn clean org.jacoco:jacoco-maven-plugin:prepare-agent package sonar:sonar \
...
+  ```bash
+  mvn initialize artifact:check-buildplan
+  ```
+* Verify that the build produces excatly the same artifacts:
+  ```bash
+  mvn -T 1C clean install -DskipTests
+  mvn -T 1C clean verify artifact:compare -DskipTests
+  ```
+
+The build will use the last Git commit timestamp as timestamp for files in `.jar` archives.
```
redcatbear, line 195: :+1: Great!

redcatbear (approved):

PR #418, opened 2024-06-21
#417: Allow numbers after dots in specificaiton item names

### LINKED ISSUE #417: Requirement item id name with dot and numbers
# Description
Following https://github.com/itsallcode/openfasttrace/blob/main/doc/user_guide.md#specification-item-id it states that

> The name part of the ID must be a character string consisting of Unicode letters and or numbers separated by underscore ("_"), hyphen ("-") or dot ("."). Whitespaces are not allowed.

We use IDs like "SR.AB.1.1.1" resulting in an item ID of `req~SR.AB.1.1.1~1`. But that gets ignored if I try to trace. If I replace all but the first dot the ID is traced again e.g. `req~SR.AB_1_1_1~1`. If I replace any other dot it will be ignored again e.g. the second dot `req~SR_AB.1_1_1~1`.

What does work is: If I have no number following a dot: `req~SR.AB.A1.B1.C1~1`. So that seems to be the issue here. No number directly after a dot.

# Environment
- OFT: 4.0.0
- OS: Windows 10
- Java Version: 22

kaklakariada:
Closes #417

### DIFF
--- a/doc/user_guide.md
+++ b/doc/user_guide.md
@@ -79,11 +79,11 @@ We use this term to better distinguish between the accepted use of the word "req
 The identifier (ID) of a [specification item](#specification-item) is a project-globally unique key which is used to refer to a specification item.
 
 The specification item ID consists of the following parts:
-* [Artifact type](#artifact-type)
-* Name
+* [Artifact type](#specification-item-artifact-type)
+* [Name](#specification-item-name)
 * [Revision](#specification-item-revision)
 
-All parts are integral to the ID. The name alone is neither unique nor complete. In OFT's native document formats the ID is represented as a character string where the three parts are separated by the tilde ("~") symbol.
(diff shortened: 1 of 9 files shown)


doc/user_guide.md
```diff
@@ -116,14 +114,30 @@ While not enforced by OFT the following strings are well established:
...
+
+We recommend using a dot `.` to create a hierarchy of items:
+
+    exporter.html5.folding
+    exporter.html5.colors
+    exporter.csv.column_names
+
+##### Specification Item Revision
+
+The revision number of a specification item is a positive integer number that can be started at zero but out of convention usually is started at one.
```
redcatbear, line 138:
```suggestion
The revision number of a specification item is a positive integer number that can be started at zero but by convention usually is started at one.
```

redcatbear (approved):

PR #445, opened 2025-03-08
Added support for requirement refinement

### LINKED ISSUE #444: Requirement refinement
As a *requirement engineer*
I want to refine requirements by linking a set of more precise requirements to an course grained requirements

in order to refine the course grained requirements. The refined requirements need to be able to fully cover the course grained requirements. The refined requirements need to be covered by needs types while the course grained requirements is fully covered by the refined requirements while the needs attribute is ignored.

Background:
Architectures of systems with multiple components often define cross component requirements the describe patterns that all components need to adhere but the interpretation of the pattern is then component specific and needs to also be implemented and tested component specific.

poldi2015:
`LinkedSpecificationItem` is covered when covered by at least one ` SpecificationItem`  of the same type.

### DIFF
--- a/api/src/main/java/org/itsallcode/openfasttrace/api/core/LinkedSpecificationItem.java
+++ b/api/src/main/java/org/itsallcode/openfasttrace/api/core/LinkedSpecificationItem.java
@@ -17,6 +17,8 @@ public class LinkedSpecificationItem
     private final Set<String> coveredArtifactTypes = new HashSet<>();
     private final Set<String> coveredArtifactTypesFromApprovedItems = new HashSet<>();
     private final Set<String> overCoveredArtifactTypes = new HashSet<>();
+    private boolean isRefined = false;
+    private boolean isRefinedApproved = false;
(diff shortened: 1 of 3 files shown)


poldi2015:
> To be frank, I don't think this is a good idea. If a requirement has a certain mandatory coverage artifact type, I find it misleading to allow covering it by a different type. Also, refinement can simply be done by covering the parent with multiple requirements on the current level.
> 
> Let us discuss the rationale behind this. As it stands, I reject the PR.

Yes let's discuss. Most likely it needs a bit more background. This is in fact one of the topics that I want to discuss with you.

What do you mean with "refinement can simply be done by covering the parent with multiple requirements on the current level.".

What I actually need is more or less: If have a requirement of type "req" withneeds is   req or (design and test). We currently use that a lot in our architecture (at least > 500 requirements) as we have two levels of architects: system and team architects. In some cases the team architects decide to refine the system level requirements, in most cases not.

PR #451, opened 2025-05-18
#449: Needs parsed past end of section

### LINKED ISSUE #449: Parsing Feature Needs from MD goes past next field
# Description
Markdown file contains a feature, with a title then a feature name, then "Needs:" then "Status:" then "Description:".
In the description there is a bulleted list.
When parsing this, OFT takes the last word of each bulleted list entry and adds it to the "Needs:".

## Steps to Reproduce
Here is an MD snippet:
```

Requirement 10100: Bridge Interrupt Outputs
-------------------------------------------------------------------------------------------
`feat~calico.10100.bridge-interrupt-outputs~1`

Needs: arch

Status:
approved

Description:

Calico shall have interrupt outputs.

Uses are all dependent on the underlying features being supported by the Calico version:

* transaction support errors
* security violations
* RAS / parity errors
* PMon threshold crossing
```

Run OFT and look at the report.html
java -jar /openfasttrace-4.1.0.jar trace -l FINE -o html -f /output/report.html /sources

Image

## Expected behavior
Stop adding to "Needs:" after the next field starts (in this case "Status:" is the next field)

# Environment
- OFT 4.1.0
- OS: Linux running on Docker image eclipse-temurin:23.0.2_7-jre-alpine-3.21
- Java Version: openjdk 23.0.2 2025-01-21
- Maven Version (in case of build problems only): n/a, downloaded prebuild jar for OFT

## Additional context
Have an easy workaround, just putting the "Needs:" at the end of the Markdown section.  But still, this seems like pretty wrong behaviour

redcatbear:
Fixed a bug where the Markdown parser collected parts of bulleted lists to a "needs" specification that should have been finalized already.

Also added a feature to suppress parsing of certain code blocks.

Closes #449.


importer/lightweightmarkup/src/main/java/org/itsallcode/openfasttrace/importer/lightweightmarkup/statemachine/LineParserState.java
```diff
@@ -23,8 +23,10 @@ public enum LineParserState
     RATIONALE,
     /** Inside a comment section */
     COMMENT,
-    /** Inside a section defining the required coverage */
-    NEEDS,
+    /** Inside a section defining the required coverage (inline form) */
+    NEEDS_LINE,
+    /** Required coverage (list form */
```
kaklakariada, line 28:
```suggestion
    /** Required coverage (list form) */
```

importer/restructuredtext/src/test/java/org/itsallcode/openfasttrace/importer/restructuredtext/TestRstSectionTitlePattern.java
```diff
@@ -85,9 +85,11 @@ void test(final String line, final String nextLine, final String expected)
         }
         else
         {
-            assertAll(() -> assertThat(
-                    "Lines '" + line + "' + '" + nextLine + "' should be recognized as a section title",
-                    result.isPresent(), is(true)), () -> assertThat(result.get().get(0), is(expected)));
+            if(result.isPresent()) {
+                assertThat(result.get().get(0), is(expected));
```
kaklakariada, line 89: Also assert that the list has exactly one item

kaklakariada (changes requested):

kaklakariada (approved):

Issue #429, opened 2024-08-13
Importing XML files with DOCTYPE fails


kaklakariada:
Importing the following file fails:

```xml
<?xml version="1.0" encoding="UTF-8" ?>
<!DOCTYPE configuration>

<configuration>
  <import class="ch.qos.logback.classic.encoder.PatternLayoutEncoder"/>
  <import class="ch.qos.logback.core.ConsoleAppender"/>

  <appender name="STDOUT" class="ConsoleAppender">
    <encoder class="PatternLayoutEncoder">
      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} -%kvp- %msg%n</pattern>
    </encoder>
  </appender>

  <root level="info">
    <appender-ref ref="STDOUT"/>
  </root>
</configuration>
```

PR #469, opened 2025-09-07
#429: Added a peek function to the factory that decides whether .xml files are ReqM2 files

redcatbear:


importer/specobject/src/main/java/org/itsallcode/openfasttrace/importer/specobject/SpecobjectImporterFactory.java
```diff
@@ -1,14 +1,22 @@
...
+
 import org.itsallcode.openfasttrace.api.importer.*;
 import org.itsallcode.openfasttrace.api.importer.input.InputFile;
 import org.itsallcode.openfasttrace.importer.xmlparser.XmlParserFactory;
 
 /**
- * An {@link ImporterFactory} for XML specobject files.
+ * An {@link ImporterFactory} for ReqM2/SpecObject XML files.
  */
 public class SpecobjectImporterFactory extends RegexMatchingImporterFactory
```
kaklakariada, line 15: Now that we have some custom logic in `supportsFile()`, we should extend `ImporterFactory` instead of `RegexMatchingImporterFactory`.

redcatbear: Done.

PR #471, opened 2025-10-21
Add support for SystemVerilog files

eanorige:
Similar to #441, but for SystemVerilog files used in silicon chip design.

Needs changelog entry and doc update.


eanorige:
@kaklakariada Thanks, added.

kaklakariada (approved):

PR #495, opened 2026-03-08
Fixed 0% SonarCloud coverage

### LINKED ISSUE #494: 0% Code Coverage in SonarCloud (JaCoCo agent not instrumenting tests)
# Description
 The code coverage reported in SonarCloud for the project is 0%, despite having multiple tests implemented. 

After investigating the `pom.xml`, we found that the issue is not the lack of tests, but a synchronization and configuration problem with JaCoCo and Surefire/Failsafe plugins. Specifically (at pom.xml):
1. The `<argLine>` parameter in both `maven-surefire-plugin` and `maven-failsafe-plugin` is being manually overwritten (to configure Mockito and Java modules), which accidentally removes the JaCoCo agent from the JVM during test execution.
2. The `prepare-agent` goal for JaCoCo is not strictly bound to the `initialize` phase.
3. The XML report format is not explicitly configured, making it difficult for SonarCloud to read the data.

Because of this, the `jacoco.exec` file is not generated, resulting in 0% coverage.

## Steps to Reproduce
1. Clone the repository.
2. Run the command `mvn clean verify`.
3. Check the `target` folder. The `jacoco.exec` file is either missing or empty, and no valid `jacoco.xml` is generated for SonarCloud.

## Expected behavior
When running `mvn clean verify`, JaCoCo should properly instrument the tests, generate the `jacoco.exec` binary, and format it into a `jacoco.xml` report. SonarCloud should then be able to read this file and display the real code coverage (which should be around 86%).

# Environment
- OFT: [4.2.2]
- OS: [ Windows 11]
- Java Version: [openjdk version "25.0.2"]
- Maven Version: [3.9.12]

## Additional context
We have already investigated and fixed this issue locally by injecting the `${jacoco.agent}` property into the `<argLine>` and adjusting the `pom.xml` configuration. We will open a Pull Request shortly linking to this issue with the proposed solution.

GloriaCP:
**### Fixes #494**
# Description
 The code coverage reported in SonarCloud for the project is 0%, despite having multiple tests implemented. 

After investigating the `pom.xml`, we found that the issue is not the lack of tests, but a synchronization and configuration problem with JaCoCo and Surefire/Failsafe plugins. Specifically (at pom.xml):
1. The `<argLine>` parameter in both `maven-surefire-plugin` and `maven-failsafe-plugin` is being manually overwritten (to configure Mockito and Java modules), which accidentally removes the JaCoCo agent from the JVM during test execution.
2. The `prepare-agent` goal for JaCoCo is not strictly bound to the `initialize` phase.
3. The XML report format is not explicitly configured, making it difficult for SonarCloud to read the data.

Because of this, the `jacoco.exec` file is not generated, resulting in 0% coverage.

## Steps to Reproduce
1. Clone the repository.
2. Run the command `mvn clean verify`.
3. Check the `target` folder. The `jacoco.exec` file is either missing or empty, and no valid `jacoco.xml` is generated for SonarCloud.

## Expected behavior
When running `mvn clean verify`, JaCoCo should properly instrument the tests, generate the `jacoco.exec` binary, and format it into a `jacoco.xml` report. SonarCloud should then be able to read this file and display the real code coverage (which should be around 86%).

(Here is the link to our SonarCloud analysis. Proved that our changes worked. 
https://sonarcloud.io/project/overview?id=gloce_openfasttrace )
# Environment
- OFT: [4.2.2]
- OS: [ Windows 11]
- Java Version: [openjdk version "25.0.2"]
- Maven Version: [3.9.12]


parent/pom.xml
```diff
@@ -15,6 +15,17 @@
         <junit.version>6.1.0-M1</junit.version>
         <junit.version>6.0.2</junit.version>
         <maven.surefire.version>3.5.4</maven.surefire.version>
+        <!-- explicit version used by jacoco agent property calc below -->
```
redcatbear, line 19:
```suggestion
        <!-- Explicit version used by JaCoCo agent property calculation below: -->
```

parent/pom.xml
```diff
@@ -15,6 +15,17 @@
...
+        <jacoco.version>0.8.14</jacoco.version>
+        <!-- compute agent argument up front so that ${jacoco.agent} is available
+             when Surefire expands its argLine (property interpolation happens before
+             plugin goals run). using ${user.home} instead of ${settings.localRepository}
+             because the latter is not yet available during property interpolation.
+             Use relative path for destfile to avoid issues with spaces in directory names. -->
```
redcatbear, line 25:
```suggestion
        <!-- Compute the agent argument up front so that ${jacoco.agent} is available
             when Surefire expands its argLine (property interpolation happens before
             plugin goals run). Using ${user.home} instead of ${settings.localRepository}
             because the latter is not yet available during property interpolation.
             Use relative path for destfile to avoid issues with spaces in directory names. -->
```

parent/pom.xml
```diff
@@ -296,8 +307,11 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
                 <configuration>
-                    <source>${java.version}</source>
-                    <target>${java.version}</target>
+                    <!-- use the release option rather than source/target so that the compiler
+                         automatically sets the system module path appropriately when the
+                         running JDK is newer than the target version. this avoids the
+                         "location of system modules is not set" error seen with JDK 25. -->
```
redcatbear, line 315:
```suggestion
                    <!-- Use the release option rather than source / target so that the compiler
                         automatically sets the system module path appropriately when the
                         running JDK is newer than the target version. This avoids the
                         "location of system modules is not set" error seen with JDK 25. -->
```

parent/pom.xml
```diff
@@ -399,8 +425,12 @@
                             ${basedir}/src/test/resources/logging.properties</java.util.logging.config.file>
                     </systemPropertyVariables>
                     <useModulePath>true</useModulePath>
-                    <argLine>
-                        <!-- Required for Java >= 18 --> ${test.args} -javaagent:${org.mockito:mockito-core:jar}</argLine>
+                    <!-- include JaCoCo agent along with the Mockito agent and any extra test
+                         arguments. the jacoco:prepare-agent goal will populate the
+                         ${jacoco.agent} property when invoked. without this the JVM that runs
+                         the tests is never instrumented and no .exec file is generated,
+                         which is exactly why SonarCloud sees 0% coverage. -->
```
redcatbear, line 434:
```suggestion
                    <!-- Include the JaCoCo agent along with the Mockito agent and any extra
                         test arguments. The jacoco:prepare-agent goal will populate the
                         ${jacoco.agent} property when invoked. Without this the JVM that runs
                         the tests is never instrumented and no .exec file is generated,
                         which is exactly why SonarCloud sees 0% coverage. -->
```

parent/pom.xml
```diff
@@ -414,7 +444,9 @@
                         <test.coverage>true</test.coverage>
                     </systemPropertyVariables>
                     <useModulePath>true</useModulePath>
-                    <argLine><!-- Required for Java >= 18 --> ${test.args} -javaagent:${org.mockito:mockito-core:jar}</argLine>
+                    <!-- same change for integration tests: we need the jacoco agent on the
+                         forked JVMs that run failsafe. -->
+                    <argLine>${jacoco.agent} ${test.args} -javaagent:${org.mockito:mockito-core:jar}</argLine>
```
redcatbear, line 451:
```suggestion
                    <!-- See in the unit tests above for a an explanation. In short, we need the JaCoCo agent on the
                         forked JVMs that run Maven Failsafe. -->
```

redcatbear (changes requested):
Thanks a lot for the valuable contribution!

I reviewed code and documentation and found a few small things that I would like to ask you to adapt in order to be in the same style as the rest of the project.

Please also add an entry to the change log for the next version.

parent/pom.xml
```diff
@@ -13,8 +13,21 @@
...
         <junit.version>6.0.3</junit.version>
         <maven.surefire.version>3.5.5</maven.surefire.version>
+        <!-- explicit version used by jacoco agent property calc below -->
+        <jacoco.version>0.8.14</jacoco.version>
+        <!-- compute agent argument up front so that ${jacoco.agent} is available
+             when Surefire expands its argLine (property interpolation happens before
+             plugin goals run). using ${user.home} instead of ${settings.localRepository}
+             because the latter is not yet available during property interpolation.
+             Use relative path for destfile to avoid issues with spaces in directory names. -->
+        <jacoco.agent>-javaagent:${user.home}\.m2\repository\org\jacoco\org.jacoco.agent\${jacoco.version}\org.jacoco.agent-${jacoco.version}-runtime.jar=destfile=jacoco.exec</jacoco.agent>
```
kaklakariada, line 26: Please use the the jacoco plugin or the dependency plugin to determine the path to the jacoco agent. This will break if Maven changes the repo layout or if the user moves the Maven repo.

testutil/.settings/org.eclipse.jdt.core.prefs
kaklakariada, line 1: Please revert these changes as this is inconsistent with other modules

kaklakariada (changes requested):

redcatbear:
Hi @GloriaCP , thanks for your contribution. @kaklakariada fixed the Sonar issue in a recent PR so I am closing this PR here.

Issue #536, opened 2026-06-02
Resolve Sonar findings of June 2026


redcatbear:
# Situation

2026-06-02 Sonar reports 375 findings. Those were collected over the years because Sonar gained new rules over time, so that previously acceptable code is now marked.

https://sonarcloud.io/project/issues?issueStatuses=OPEN&id=org.itsallcode.openfasttrace%3Aopenfasttrace-root

# Acceptance Criteria

1. All Critical (0), High (31) and Medium (70) are closed
2. (optional) All Low (262) and Info (12) are closed

redcatbear:
Done.

PR #537, opened 2026-06-02
refactoring/536_resolve-sonar-findings-of-june-2026

redcatbear:
This PR fixes a number of Sonar findings that OFT collected over the years with Sonar getting new and more tight rules.

Contributes to #536. I don't think we will be able to do everything in a single PR.


redcatbear (commented):
@kaklakariada, the review is less work than it looks. Most changes fall into on of these categories:

1. String concatenation to text block
2. Remove superfluous code (like unthrown exceptions)
3. Streams improvements (immutability in collector)
4. Resource protection

kaklakariada (approved):

PR #550, opened 2026-06-19
Refactoring/536 fix abstract class names

redcatbear:
All abstract class names now start with `Abstract`.

Contributes to #536.


doc/images/uml/class/cl_importer_overview.svg
kaklakariada, line 1: Optional: combine the three outer packages into org.itsallcode.openfasttrace to avoid clutter

kaklakariada (approved):

PR #553, opened 2026-07-01
added coverage of multiple ids in tags

Davidius86:
This PR enhances tags to define multiple covered ids in a tag.

ChatGPT 5.4 has been used for the first draft of the implementation and for the test.
The test was in almost in a good condition. The implementation has been heavily refactored after generation.

Documentation is written manually.

### DIFF
--- a/doc/changes/changes_4.6.0.md
+++ b/doc/changes/changes_4.6.0.md
@@ -19,3 +19,7 @@ We moved some GitHub action permissions from workflow-level to job-level.
 * #546: Replaced `OsDetector` with JUnit5's `EnabledOnOs` annotation.
 * #544: Replaced optional parameter with null check
 * #543: Made `CliException` a `RuntimeException`
+
+## Features
+
+* #553 Tag importer supports multiple covered ids
--- a/doc/changes/changes_4.6.0.md
+++ b/doc/changes/changes_4.6.0.md
@@ -22,4 +22,4 @@ We moved some GitHub action permissions from workflow-level to job-level.
 
 ## Features
 
-* #553 Tag importer supports multiple covered ids
+* #553 Tag importer supports multiple covered IDs
--- a/doc/user_guide.md
+++ b/doc/user_guide.md
@@ -708,7 +708,7 @@ To avoid conflict with the formats actual contents, you embed these definitions 
 Tags have the following format:
 
 ```
-[ <covered-artifact-type> -> <specification-object-id> ]
+[ <covered-artifact-type> -> <list-of-specification-object-ids> ]
 ```
 
 Spaces above were only added for readability. They are optional. In fact usually people prefer a more compact form.
@@ -737,7 +737,7 @@ Examples:
 // [impl~validate-password~2->dsn~validate-authentication-request~1]
 ```
 
-##### Forwarding Requirements
+##### Needed Coverage
 
 When using UML models as design document files like UML models it is useful to add needed coverage as well. To do this, you can use the following format:
 
--- a/importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
+++ b/importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
@@ -2,7 +2,9 @@ package org.itsallcode.openfasttrace.importer.tag;
 
 import static java.util.Collections.emptyList;
 
-import java.util.*;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
 import java.util.function.Predicate;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
@@ -23,18 +25,21 @@ class LongTagImportingLineConsumer extends AbstractRegexLineConsumer
     private static final String OPTIONAL_WHITESPACE = "\\s*";
     private static final String TAG_PREFIX = "\\[";
     private static final String TAG_SUFFIX = "\\]";
-    private static final String NEEDS_COVERAGE = ">>" + OPTIONAL_WHITESPACE + "(\\p{Alpha}+(?:"
-            + OPTIONAL_WHITESPACE
-            + "," + OPTIONAL_WHITESPACE + "\\p{Alpha}+)*)";
+    private static final String COVERED_IDS = SpecificationItemId.ID_PATTERN + "(?:"
+            + OPTIONAL_WHITESPACE + "," + OPTIONAL_WHITESPACE + SpecificationItemId.ID_PATTERN
+            + ")*";
+    private static final String NEEDS_COVERAGE = ">>" + OPTIONAL_WHITESPACE
+            + "(?<neededArtifactTypes>\\p{Alpha}+(?:" + OPTIONAL_WHITESPACE + ","
+            + OPTIONAL_WHITESPACE + "\\p{Alpha}+)*)";
     private static final String TAG_REGEX = TAG_PREFIX + OPTIONAL_WHITESPACE//
-            + "(" + COVERING_ARTIFACT_TYPE_PATTERN + ")"
+            + "(?<artifactType>" + COVERING_ARTIFACT_TYPE_PATTERN + ")"
             + "(?:" + SpecificationItemId.ARTIFACT_TYPE_SEPARATOR
             // [impl->dsn~import.full-coverage-tag-with-name-and-revision~1]
-            + "(" + SpecificationItemId.ITEM_NAME_PATTERN + ")?"
+            + "(?<customName>" + SpecificationItemId.ITEM_NAME_PATTERN + ")?"
             + SpecificationItemId.REVISION_SEPARATOR
-            + SpecificationItemId.ITEM_REVISION_PATTERN + ")?" //
+            + "(?<revision>" + SpecificationItemId.ITEM_REVISION_PATTERN + "))?" //
             + OPTIONAL_WHITESPACE + "->" + OPTIONAL_WHITESPACE //
-            + "(" + SpecificationItemId.ID_PATTERN + ")" //
+            + "(?<coveredIds>" + COVERED_IDS + ")" //
             + OPTIONAL_WHITESPACE + "(?:" + NEEDS_COVERAGE + OPTIONAL_WHITESPACE + ")?" //
             + TAG_SUFFIX;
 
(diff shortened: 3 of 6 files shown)


doc/user_guide.md
```diff
@@ -708,7 +708,7 @@ To avoid conflict with the formats actual contents, you embed these definitions
 Tags have the following format:
 
 ```
-[ <covered-artifact-type> -> <specification-object-id> ]
+[ <covered-artifact-type> -> <list-of-specification-object-ids> ]
```
redcatbear, line 711:
```suggestion
[ <covered-artifact-type> -> <specification-object-id>]

If a piece of code covers multiple specification items, you can also list them separated by comma.

[ <covered-artifact-type> -> <specification-object-id>,  <specification-object-id>]
```

redcatbear: Thanks @Davidius86 for the PR. I am started reviewing it. I will update main to 2.6.0 and create the release letter in another PR, so that we have a place where we can mention the change.

redcatbear: Completeness: Please also add a short example of the syntax to `.agents/skills/openfasttrace/SKILL.md`.

redcatbear: Completeness: Please add an short entry for this in the change log too (we are currently working towards 4.6.0).

doc/user_guide.md
```diff
@@ -737,7 +737,7 @@ Examples:
 // [impl~validate-password~2->dsn~validate-authentication-request~1]
 ```
 
-##### Forwarding Requirements
+##### Needed Coverage
```
redcatbear, line 740: Good find. 👍 

Compliance: I am guessing you are using an LLM to help with the work. Please note that according to our contribution guideline, LLM-assisted PR must state that.

importer/tag/src/test/java/org/itsallcode/openfasttrace/importer/tag/TestTagImporter.java
redcatbear, line 1: Completeness: Please also add tests that the features with requiring coverage and partial spec item definition still work in combination with the comma-separated covered IDs.

importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -53,18 +58,27 @@ public void processMatch(final Matcher matcher, final int lineNumber, final int
...
-        this.listener.addCoveredId(coveredId);
-        neededArtifactTypes.forEach(listener::addNeededArtifactType);
+        coveredIds.forEach(this.listener::addCoveredId);
+        neededArtifactTypes.forEach(this.listener::addNeededArtifactType);
         this.listener.endSpecificationItem();
     }
 
-    private static List<String> parseCommaSeparatedList(final String input)
+    private static List<SpecificationItemId> parseCoveredIds(final String input)
+    {
```
redcatbear, line 93: Safety: Add null handling, please.

importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -112,21 +126,32 @@ private static int parseRevision(final String revision)
...
-        return coveredId.getName();
+        return joinCoveredIdNamesWithHyphen(coveredIds);
     }
 
-    private String generateUniqueName(final SpecificationItemId coveredId, final int lineNumber,
+    private String generateUniqueName(final List<SpecificationItemId> coveredIds, final int lineNumber,
             final int counter)
     {
-        final String uniqueName = this.file.getPath() + lineNumber + counter + coveredId;
+        final String CoveredIdStringsJoinedWithHyphen = coveredIds.stream()
```
redcatbear: UX: This is where the compactness gets to be a problem. While it is nice to be able to cover multiple spec ids in one line, the generated single ID from the code gets excessively long. I would suggest to instead create one "virtual" covering spec item id per covered id instead of a really long one that won't be readable.

redcatbear (changes requested):

importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
Davidius86: I'm not sure whether I got this correctly.
Does "virtual" mean that multiple new specification ids are generated, one per covered id?
That would create one entry in the trace report for each covered id of this tag.
It would also require to cover each of the specification ids derived from the covered ids.

I have only a limited knowledge on the use cases but I think I would be surprised if a tag would create more than one specification id. Especially if this behavior changes depending on whether a name has been specified or is auto generated.

You have a valid point here, the readability of the concatenated name is definitely low. The following options came to my mind:
1. Use only the name of the first covered id (hides information / may be misinterpret)
2. Use only a hash without any naming (decreases readability)
3. Use an abbreviation (decreases readability)

Sorry, for this lengthy post. Please tell me which solution you prefer and I will adapt the PR accordingly.

redcatbear: In a way a conceptual issue surfaces here. Usually requirements span a more or less tree-like structure that branches out from abstract to concrete. Where a top-level feature is the most abstract and the implementation the most concrete.

So it is rare that one code place covers a lot of higher-level requirements. At least if you write short methods and keep the coverage tags in the smallest possible scope.

So normally a method should cover only a single design requirements. In rare cases two or three, but never more.

This PR adds syntactic sugar that makes the tag more compact. But it is limited for example by newlines because then the simple parser won't work anymore. Especially, since you cannot predict what the comment characters are that need to come after line breaks.

All that is to say that the best solution here is to generate one item per covered ID. The only non-confusing exception would be if instead of just the covered artifact type there was the full tag ID. But we don't want to over-complicate things.

Let me make an example to make this more concrete:

```Java
// [impl -> dsn~assign-admin-privilege~1, dsn~log-security-event~3]
public void assignAdminPrivilege(final User user)
```

Should become

1. `impl~assign-admin-privilege-1b2f456~0` → `dsn~assign-admin-privilege~1`
2. `impl~log-security-event-9876a3~0` → `dsn~log-security-event~3`

redcatbear: And

```
# [dsn -> req~assign-admin-privilege~1, req~log-security-event~3 >> impl, utest]
@startuml
```

Should become


1. `dsn~assign-admin-privilege-1b2f456~0` → `req~assign-admin-privilege~1` (needs: `impl`, `utest`)
2. `impl~log-security-event-9876a3~0` → `dsn~log-security-event~3` (needs: `impl`, `utest`)

Which will probably automatically work, but needs a test at least.

importer/tag/src/test/java/org/itsallcode/openfasttrace/importer/tag/TestTagImporter.java
redcatbear: That test looks good now. I think I will have to adapt the IDE plugin once this is merged. :-)

importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -53,18 +58,32 @@ public void processMatch(final Matcher matcher, final int lineNumber, final int
...
 
-    private static List<String> parseCommaSeparatedList(final String input)
+    private static List<SpecificationItemId> parseCoveredIds(final String input)
+    {
+        if (input == null)
+        {
+            return emptyList();
+        }
+
+        return Arrays.stream(input.split(","))
```
redcatbear, line 98:
```suggestion
        }
        return Arrays.stream(input.split(","))
```

importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -78,28 +97,28 @@ private static List<String> parseCommaSeparatedList(final String input)
...
 
-    private void logItem(final int lineNumber, final SpecificationItemId coveredId,
+    private void logItem(final int lineNumber, final List<SpecificationItemId> coveredIds,
             final List<String> neededArtifactTypes, final SpecificationItemId generatedId)
     {
         if (neededArtifactTypes.isEmpty())
         {
             LOG.finest(() -> "File " + this.file + ":" + lineNumber + ": found '" + generatedId
-                    + "' covering id '" + coveredId);
+                    + "' covering ids " + coveredIds);
```
redcatbear:
```suggestion
                    + "' covering IDs " + coveredIds);
```

importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
```diff
@@ -78,28 +97,28 @@ private static List<String> parseCommaSeparatedList(final String input)
...
         }
         else
         {
             LOG.finest(() -> "File " + this.file + ":" + lineNumber + ": found '" + generatedId
-                    + "' covering id '" + coveredId + "', needs artifact types "
+                    + "' covering ids " + coveredIds + ", needs artifact types "
```
redcatbear:
```suggestion
                    + "' covering IDs " + coveredIds + ", needs artifact types "
```

doc/changes/changes_4.6.0.md
```diff
@@ -18,3 +18,7 @@ We moved some GitHub action permissions from workflow-level to job-level.
 * #546: Replaced `OsDetector` with JUnit5's `EnabledOnOs` annotation.
 * #544: Replaced optional parameter with null check
 * #543: Made `CliException` a `RuntimeException`
+
+## Features
+
+* #553 Tag importer supports multiple covered ids
```
redcatbear:
```suggestion
* #553 Tag importer supports multiple covered IDs
```

importer/tag/src/main/java/org/itsallcode/openfasttrace/importer/tag/LongTagImportingLineConsumer.java
redcatbear: Your new tests look like this is going in the right direction already.

Davidius86: Sure, will do.

Side note: Thank you for sharing this perspective; it gave me a much clearer understanding of the problem we're facing. I do think, though, that in practice requirements often arise from different viewpoints (safety, security, user acceptance, technical limitations, architectural constraints, ... ) instead of following a purely tree-like structure.

redcatbear: You are right. Especially cross-cutting requirements can lead to multiple requirements being covered by one code block. If it is more than three though, this change will produce a really long line. Anyway, it is a complementary notation, so the old one is not affected. Should be fine.

redcatbear:
@Davidius86 the PR breaks the self-trace. That is not uncommon when requirement IDs show up in tests. You can either use string concatenation to fix this or comments with `oft:off` and `oft:on` to block certain lines from being scanned.

redcatbear (changes requested):

redcatbear (approved):

PR #560, opened 2026-07-11
Add support for filtering by status

### LINKED ISSUE #519: Support filtering specification items by status
OpenFastTrace currently supports filtering imported specification items by artifact type and tags, but not by item status.

**Use case:**
Projects can mark reverse-engineered or not-yet-reviewed items with the conventional field `Status: draft`. Those items should be excluded from trace verification until promoted, without having to duplicate the state in `Tags: draft` or require an unrelated `Tags: active` marker.

**Requested behavior:**
- Add a trace/import filter for status values, for example include only non-draft items or exclude `Status: draft`.
- Keep the existing tag filtering behavior unchanged.
- Document how status filtering interacts with items that do not declare a status.

**Workaround:**
- Filter by status `active`
- Add `Tags: draft` to items in status draft
- Replace tag `draft` with `active` once an item is no long a draft.

redcatbear:
This pull request adds functionality to filter items by their status.
Introduces new command line switch `-w` and `--wanted-statuses`

Closes #519.

### DIFF
--- a/api/src/main/java/org/itsallcode/openfasttrace/api/FilterSettings.java
+++ b/api/src/main/java/org/itsallcode/openfasttrace/api/FilterSettings.java
@@ -1,7 +1,10 @@
 package org.itsallcode.openfasttrace.api;
 
+import java.util.EnumSet;
 import java.util.Objects;
 import java.util.Set;
+
+import org.itsallcode.openfasttrace.api.core.ItemStatus;
 
 /**
  * Settings for import filtering
@@ -9,12 +12,14 @@ import java.util.Set;
 public final class FilterSettings
 {
     private final Set<String> artifactTypes;
+    private final Set<ItemStatus> wantedStatuses;
     private final Set<String> tags;
     private final boolean withoutTags;
 
     private FilterSettings(final Builder builder)
     {
         this.artifactTypes = builder.artifactTypes;
+        this.wantedStatuses = builder.wantedStatuses;
         this.tags = builder.tags;
         this.withoutTags = builder.withoutTags;
     }
@@ -27,6 +32,16 @@ public final class FilterSettings
     public Set<String> getArtifactTypes()
     {
         return Set.copyOf(this.artifactTypes);
+    }
+
+    /**
+     * Get the statuses the filter must match.
+     * 
+     * @return statuses that must be matched
+     */
+    public Set<ItemStatus> getWantedStatuses()
+    {
+        return Set.copyOf(this.wantedStatuses);
     }
 
     /**
@@ -61,6 +76,16 @@ public final class FilterSettings
     }
 
     /**
+     * Check if the status filter is set.
+     * 
+     * @return {@code true} if the status filter is set
+     */
+    public boolean isStatusCriteriaSet()
+    {
+        return this.wantedStatuses != null && !this.wantedStatuses.isEmpty();
+    }
+
+    /**
      * Check if the tag filter is set.
      * 
      * @return {@code true} if the tag filter is set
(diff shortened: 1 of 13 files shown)


api/src/main/java/org/itsallcode/openfasttrace/api/FilterSettings.java
```diff
@@ -60,6 +75,16 @@ public boolean isArtifactTypeCriteriaSet()
...
     }
 
+    /**
+     * Check if the status filter is set.
+     * 
+     * @return {@code true} if the status filter is set
+     */
+    public boolean isStatusCriteriaSet()
+    {
+        return this.wantedStatuses != null && !this.wantedStatuses.isEmpty();
```
kaklakariada, line 85: Add a non null check to the constructor and remove the null Check here

api/src/test/java/org/itsallcode/openfasttrace/api/importer/TestFilterSettings.java
```diff
@@ -31,6 +32,17 @@ void testBuilder()
...
 
+    @Test
+    void testBuilderWithStatuses()
+    {
+        final ItemStatus[] expectedStatuses = { ItemStatus.APPROVED, ItemStatus.DRAFT };
+        final FilterSettings filterSettings = FilterSettings.builder() //
+                .wantedStatuses(new HashSet<>(Arrays.asList(expectedStatuses))) //
+                .build();
+        assertThat(filterSettings.getWantedStatuses(), containsInAnyOrder(expectedStatuses));
+        assertFilterSet(filterSettings, true);
```
kaklakariada, line 43: Use assertAll

doc/changes/changes_4.6.0.md
```diff
@@ -1,13 +1,20 @@
...
-Code name: ??
-
-We also updated test and build dependencies to fix vulnerabilities. Runtime code is not affected, so no update is required.
+Code name: Status Filter at Import
 
 ## Summary
 
-We moved some GitHub action permissions from workflow-level to job-level.
+We added a new feature to filter specification items by status at import time via `-w` or `--wanted-statuses` CLI parameters.
+This is helpful when your project uses specification documents for planning future requirements (marked by the status "draft" or "porposed").
```
kaklakariada, line 8: A typical use case is to exclude only `draft` items. If I understand correctly, this would require specifying all other status in this option?

Another remark: We often don't use the status field for items. Only when we add items we mark them as draft. Would it be possible to allow including items without status?

product/src/test/java/org/itsallcode/openfasttrace/cli/CliStarterInternalIT.java
```diff
@@ -305,6 +306,41 @@ void testTraceWithFilteredArtifactType() {
...
+                OUTPUT_FILE_PARAMETER, this.outputFile.toString(),
+                WANTED_STATUSES_PARAMETER, "draft, proposed",
+                REPORT_VERBOSITY_PARAMETER, "all");
+        assertAll(
+            () -> assertThat(status, equalTo(ExitStatus.OK)),
+            () -> assertOutputFileExists(true),
+            () -> assertThat(getOutputFileContent(), containsString("2 total")),
+            () -> assertThat(getOutputFileContent(), containsString("req~draft~1")),
+            () -> assertThat(getOutputFileContent(), containsString("req~proposed~1")),
+            () -> assertThat(getOutputFileContent(), not(containsString("req~approved~1")))
```
kaklakariada:
```suggestion
            () -> assertThat(getOutputFileContent(), not(containsString("approved")))
```
be more strict

kaklakariada (changes requested):

kaklakariada (approved):

PR #568, opened 2026-08-01
feature/251_colored--transitive-failiure

redcatbear:
Exposed transitive defects via the `LinkedSpecificationItem` API.

Now rendering HTML and plain text report to make the distinction between direct and transitive defects.


reporter/plaintext/src/main/java/org/itsallcode/openfasttrace/report/plaintext/PlainTextReport.java
```diff
@@ -108,10 +110,34 @@ private void renderResultStatus(final PrintStream report)
...
+        }
+        if (item.isTransitiveDefect())
+        {
+            return this.formatter.formatTransitiveNotOk("not ok (transitive)");
+        }
+        return this.formatter.formatNotOk(NOT_OK);
     }
 
     // [impl->dsn~reporting.plain-text.summary~2]
+    // [impl->dsn~reporting.plain-text.summary-line~1]
```
redcatbear: The two requirement IDs a not speaking enough.

kaklakariada (approved):

PR #580, opened 2026-08-23
#521: Draft: JSON-based exchange format for OFT items and reports

### LINKED ISSUE #521: Draft: JSON-based exchange format for OFT items and reports
#### Description
We propose a JSON-based exchange format for OpenFastTrace (OFT) specification items and reports. This format is intended to support both exports (unlinked items) and reports (linked/traced items) while ensuring bit-identical reproducibility when desired.

#### Key Features
- **Header Section**: Contains metadata including schema version, OFT version, parameters, and a `mode` marker (`export` or `report`).
- **Specification Items**: A flat list of items using exact OFT terminology. The `covering` list is simplified to only include target IDs.
- **Tracing Details**: An optional section per item (for `report` mode) containing linkage status, coverage results, and defect markers.
- **Report Summary**: An optional global summary for traced results.
- **Reproducibility**: The `timestamp` is optional. For bit-identical output, it should be omitted, and all lists/keys should be sorted.

#### Draft Format Example (Report Mode)

```json
{
  "header": {
    "schema_version": "1.0",
    "oft_version": "4.5.0",
    "mode": "report",
    "timestamp": "2026-05-24T20:09:00Z",
    "parameters": {
      "input_paths": ["src/main/resources/specs"],
      "output_format": "json",
      "wanted_artifact_types": ["req", "dsn", "impl"],
      "show_origin": true
    }
  },
  "specification_items": [
    {
      "id": "req-data-storage-1",
      "version": 1,
      "doctype": "req",
      "shortdesc": "Data Storage Requirement",
      "status": "approved",
      "sourcefile": "specs/requirements.md",
      "sourceline": 42,
      "description": "The system shall store data in a JSON format.",
      "coverage": {
        "needed_coverage_types": ["dsn"]
      },
      "covering": [
        { "id": "feat-json-export-1" }
      ],
      "tracing_details": {
        "is_defect": false,
        "shallow_coverage_status": "COVERED",
        "deep_coverage_status": "COVERED",
        "covered_artifact_types": ["dsn"],
        "uncovered_artifact_types": [],
        "links": [
          {
            "target_id": "dsn-json-storage-1",
            "target_version": 1,
            "target_doctype": "dsn",
            "status": "covered shallow",
            "direction": "incoming"
          }
        ]
      }
    }
  ],
  "report_summary": {
    "result_status": "OK",
    "total_item_count": 1,
    "defect_item_count": 0,
    "coverage_statistics": {
      "fully_covered": 1,
      "partially_covered": 0,
      "uncovered": 0
    }
  }
}
```

#### Reproducibility Guidelines
1. Omit `header.timestamp`.
2. Sort `specification_items` by `id` then `version`.
3. Sort all nested lists (e.g., `links`, `covering`, `covered_artifact_types`).
4. Sort keys alphabetically within all objects.

redcatbear:
Discussion draft for the OFT's own native document interchange and report format.

Closes #521.


doc/schemas/openfasttrace.schema.json
```diff
@@ -0,0 +1,574 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "https://github.com/itsallcode/openfasttrace/blob/main/doc/schemas/openfasttrace.schema.json",
+  "title": "OpenFastTrace JSON formats",
+  "description": "Schema for OpenFastTrace exchange (OFTX) and report (OFTR) documents. Properties whose value equals the declared default are omitted from serialized documents.",
```
kaklakariada, line 5: Just an idea: this mixes exchange and reports. Would it be clearer to have separate schemas?

I guess it's less duplication to leave it like this.

doc/schemas/openfasttrace.schema.json
```diff
@@ -0,0 +1,574 @@
...
+          "enum": [
+            "exchange",
+            "report"
+          ]
+        },
+        "reproducible": {
+          "type": "boolean",
+          "default": false,
+          "description": "Whether running OFT again with the recorded parameters can reproduce the document bit-identically."
+        },
```
kaklakariada, line 80: Do we need this? When is a report not reproducible? Only when the order in which files are read is different. But then OFT should sort the report reproducibly.

doc/schemas/openfasttrace.schema.json
```diff
@@ -0,0 +1,574 @@
...
+            "file",
+            "directory"
+          ]
+        },
+        "reference": {
+          "default": "relative",
+          "enum": [
+            "absolute",
+            "relative"
+          ]
```
kaklakariada: Does this mean that when the user specifies a relative path, the report contains only relative paths starting from that path, and when they specify an absolute path, all files under this path will be absolute in this report?

doc/schemas/openfasttrace.schema.json
```diff
@@ -0,0 +1,574 @@
...
+      },
+      "additionalProperties": false
+    },
+    "output": {
+      "type": "object",
+      "properties": {
+        "format": {
+          "enum": [
+            "oftx",
+            "oftr"
```
kaklakariada: These constants are a bit cryptic. Could we write this out, e.g. "oft-exchange" or similar?

doc/schemas/openfasttrace.schema.json
```diff
@@ -0,0 +1,574 @@
...
+      },
+      "additionalProperties": false
+    },
+    "exchangeItem": {
+      "$ref": "#/$defs/item",
+      "properties": {
+        "trace": false
+      }
+    },
+    "summary": {
```
kaklakariada:
```suggestion
    "traceSummary": {
```
This seems to be the result of a trace, nut just some statistics about the items.

doc/schemas/openfasttrace.schema.json
```diff
@@ -0,0 +1,574 @@
...
+        },
+        "name": {
+          "$ref": "#/$defs/range"
+        },
+        "revision": {
+          "$ref": "#/$defs/range"
+        }
+      },
+      "additionalProperties": false
+    },
```
kaklakariada, line 529: Not sure if we need located IDs in the report. I would only add this if someone needs it.

kaklakariada (changes requested):
General: looks very good!

Some more comments in the schema could help with understanding the proposal.

doc/schemas/openfasttrace.schema.json
redcatbear: Order, timestamps. Some people like to have timestamps in the reports as record. Good as historical document. Less so if you need it reproducible.

doc/schemas/openfasttrace.schema.json
redcatbear: Absolute paths seldom are useful in exchange. More so in a record (e.g., as proof of a tracing run). I understand "relative" as relative to the project root. Typically a repo checkout.

doc/schemas/openfasttrace.schema.json
redcatbear: Can do. I just thought we align that with the extension `.oftx.json`. But writing it out inside the document is OK.

doc/schemas/openfasttrace.schema.json
redcatbear: Think of it this way: a JSON based report fits nicely into a JavaScript report as Poldi did. Then you want location for jump URLs.

doc/schemas/openfasttrace.schema.json
redcatbear: Yes, duplication avoidance is the main theme here. What ASPEC proved is that a machine-readable report is a list of items plus links plus verdict. That's a concept that I like and would like to keep.

doc/schemas/openfasttrace.schema.json
redcatbear: I added the optional timestamp to the header. It can only be used when `reproducible` is set to `false`.

doc/schemas/openfasttrace.schema.json
redcatbear: Changed to `oftExchange` and `oftReport` in alignement with the other enum values. Google JSON guide.

doc/schemas/openfasttrace.schema.json
redcatbear: Renamed to `relativeToWorkingDirectory` for better clarity.

kaklakariada: I think we only should use relative paths and omit this field.

kaklakariada (approved):

kaklakariada (approved):

PR #583, opened 2026-08-31
#582: Fix Markdown importer dropping items after a code block following a title

### LINKED ISSUE #582: Markdown importer silently drops all requirements after a fenced code block that directly follows a heading
### Bug description

When a fenced code block's opening fence directly follows a Markdown heading (with only blank lines in between), the importer misparses the fences and silently ignores every specification item after the block's closing fence. There is no error or warning — `trace` reports `ok - 0 total` and exits 0, and `impl->req` coverage tags pointing at the dropped items show up as orphaned.

### Minimal reproduction

`spec.md`:

````markdown
# Spec

## A diagram before any requirement

```mermaid
graph TD
  A --> B
```

## Some requirement

`req~example~1`

A requirement that should be imported.

Needs: impl
````

```
$ java -jar openfasttrace-4.9.0.jar convert -o specobject spec.md
<?xml version="1.0" encoding="UTF-8"?>
<specdocument></specdocument>

$ java -jar openfasttrace-4.9.0.jar trace spec.md
ok - 0 total
```

**Expected:** `req~example~1` is imported (1 item). **Actual:** no items, no diagnostic.

The info string is irrelevant — ` ```mermaid `, ` ```json `, ` ```foobar `, and a bare ` ``` ` all trigger it. What matters is the position: move the same block so a paragraph of text precedes it and the requirement imports fine.

### Root cause

In `MarkdownImporter.configureTransitions()`, the `TITLE` state (entered after a heading) has no transition for `MdPattern.CODE_BEGIN`. The opening fence therefore matches `TITLE → START` via `EVERYTHING`, and the *closing* ` ``` ` then matches `CODE_BEGIN` from `START` (the `\w*` info string is optional), opening a phantom `CODE_BLOCK`. Since `LineParserStateMachine` keeps its state on unmatched lines, everything up to EOF (or the next stray fence) is swallowed.

This looks like a gap in the fix for #280 (#433), which added the `CODE_BLOCK` state but only wired `CODE_BEGIN` from `START`.

### Versions

Reproduced with openfasttrace 4.9.0 (latest release), OpenJDK 21, Linux. Originally surfaced in JabRef via the `org.itsallcode.openfasttrace` Gradle plugin 3.2.0.

### Impact

Silent requirement loss: spec authors get zero signal that items vanished from tracing. JabRef currently works around it by moving diagrams after the last requirement.

I'll open a PR adding the missing `TITLE → CODE_BLOCK` transition plus a regression test.

### PR DESCRIPTION
**Pull request by koppor**
Fixes #582: a fenced code block whose opening fence directly follows a section title made the importer silently drop every specification item after the block, because the `TITLE` state had no `CODE_BEGIN` transition and the closing fence then opened a phantom code block. This adds the missing `TITLE → CODE_BLOCK` transition, a regression test (verified to fail without the fix), and a changelog entry.

AI assistance note: analysis, fix, and test were written with Claude Code and reviewed by me; `mvn -T 1C verify` (incl. self-trace) passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014Hgw7DCVCJ6hEmxYz5cU8K

### DIFF
--- a/importer/markdown/src/test/java/org/itsallcode/openfasttrace/importer/markdown/TestMarkdownMarkupImporter.java
+++ b/importer/markdown/src/test/java/org/itsallcode/openfasttrace/importer/markdown/TestMarkdownMarkupImporter.java
@@ -254,11 +254,18 @@ class TestMarkdownMarkupImporter extends AbstractLightWeightMarkupImporterTest
                 ```
 
                 `req~example~1`
+
+                `req~example~2`
                 """,
-                contains(item()
-                        .id(SpecificationItemId.parseId("req~example~1"))
-                        .location("code_block_after_title.md", 8)
-                        .build()));
+                contains(
+                        item()
+                                .id(SpecificationItemId.parseId("req~example~1"))
+                                .location("code_block_after_title.md", 8)
+                                .build(),
+                        item()
+                                .id(SpecificationItemId.parseId("req~example~2"))
+                                .location("code_block_after_title.md", 10)
+                                .build()));
     }
 
     @ParameterizedTest
--- a/importer/markdown/src/test/java/org/itsallcode/openfasttrace/importer/markdown/TestMarkdownMarkupImporter.java
+++ b/importer/markdown/src/test/java/org/itsallcode/openfasttrace/importer/markdown/TestMarkdownMarkupImporter.java
@@ -242,6 +242,25 @@ class TestMarkdownMarkupImporter extends AbstractLightWeightMarkupImporterTest
                 emptyIterable());
     }
 
+    @Test
+    void testWhenCodeBlockDirectlyFollowsTitleThenFollowingSpecificationItemMustBeDetected()
+    {
+        assertImport("code_block_after_title.md", """
+                ## A diagram before any requirement
+
+                ```mermaid
+                graph TD
+                  A --> B
+                ```
+
+                `req~example~1`
+                """,
+                contains(item()
+                        .id(SpecificationItemId.parseId("req~example~1"))
+                        .location("code_block_after_title.md", 8)
+                        .build()));
+    }
+
     @ParameterizedTest
     @CsvSource(
     {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by koppor** [comment]
Fix seems to be straight-forward. Would really help us while using OFT.

File: importer/markdown/src/test/java/org/itsallcode/openfasttrace/importer/markdown/TestMarkdownMarkupImporter.java [hunk 1]
```diff
@@ -242,6 +242,25 @@ void testWhenInsideMarkdownCodeBlockThenNoSpecificationItemMustBeDetected(final
...
+    {
+        assertImport("code_block_after_title.md", """
+                ## A diagram before any requirement
+
+                ```mermaid
+                graph TD
+                  A --> B
+                ```
+
+                `req~example~1`
```
redcatbear, line 256: I think adding a second requirement for safety here would be good to prove that the next ones are not dropped.

redcatbear (changes requested):
Please add a second requirement to the unit test.

importer/markdown/src/test/java/org/itsallcode/openfasttrace/importer/markdown/TestMarkdownMarkupImporter.java
koppor: Added `req~example~2` right after — the test now asserts both items are detected.

redcatbear: Thanks. Perfect.

redcatbear (approved):

redcatbear:
Thank you @koppor for contributing this bug fix. Very much appreciated!

koppor:
> Thank you @koppor for contributing this bug fix. Very much appreciated!

Thank you for the review and the AI-"supported" PRs. 😅🥂

PR #595, opened 2026-09-14
documentation/improved-navigation

redcatbear:
Improve page-to-page navigation.
Intentional ToC placement.
Removed individual changelog entries from side bar.


InAnYan:
Thanks for the PR! I think this is a good improvement, as the user guide was already quite big.

But on the other hand, I'm so used that this was  only a 1 file 🤣, easier to search for something and jump around

PR #598, opened 2026-09-15
feat: make SpecificationItem fully immutable

### LINKED ISSUE #572: Make SpecificationItem immutable with toBuilder()
## Goal

Make `SpecificationItem` fully immutable by replacing its public mutation API with a builder-copy workflow.

## Scope

- Remove the public `SpecificationItem.addCoveredId(SpecificationItemId)` method as a breaking API change.
- Add `SpecificationItem.toBuilder()`, returning an independent builder pre-populated with every item property, including located IDs/ranges, text fields, status, location, relations, tags, and forwarding state.
- Defensively snapshot all builder-owned collections when building a `SpecificationItem`, so later builder changes cannot affect an existing item.
- Preserve the linker’s derived covered-ID behavior: when it must add a missing covered ID, `LinkedSpecificationItem` replaces its wrapped item with `item.toBuilder().addCoveredId(...).build()` instead of mutating the original.
- Update `dsn~specification-item` to revision 4 and update corresponding implementation and unit-test trace markers.

## Acceptance Criteria

- A `toBuilder().build()` copy retains all source properties.
- Changing a copied builder produces a new item and leaves the source unchanged.
- Mutating a builder after `build()` does not change the built item’s collections.
- Linked items continue to expose inferred covered IDs after linking, while the source `SpecificationItem` remains unchanged.
- `./oft-self-trace.sh` and `mvn -T 1C verify` pass.

bkalika:
Closed #572


doc/spec/system_requirements.md
```diff
@@ -242,7 +242,7 @@ Needs: req
 ### Anatomy of Specification Items
 
 #### Specification Item
-`req~specification-item~2`
+`req~specification-item~4`
```
kaklakariada, line 245: * You didn't update the content of this item.
  * If you plan to update it: please only increment the revision by one
  * If you don't plan to update it: please don't increment the revision

This is an internal refactoring, so updating the spec is not necessary in my view.

doc/spec/design.md
```diff
@@ -864,7 +864,7 @@ Needs: impl, utest
 ### Internal Data Structures
 
 #### Specification Item
-`dsn~specification-item~3`
+`dsn~specification-item~4`
```
kaklakariada, line 867: You didn't update the text, see my comment in doc/spec/system_requirements.md

api/src/main/java/org/itsallcode/openfasttrace/api/core/LinkedSpecificationItem.java
```diff
@@ -160,7 +166,10 @@ private void addMyItemIdToCoveringItem(final LinkedSpecificationItem coveringIte
         if (coveringItem.getItem().getCoveredIds() != null
                 && !coveringItem.getItem().getCoveredIds().contains(getId()))
         {
-            coveringItem.getItem().addCoveredId(getId());
+            coveringItem.getItem()
+                .toBuilder()
+                .addCoveredId(getId())
+                .build();
```
kaklakariada, line 172: This won't solve the issue. Calling `build()` will just create a new instance, but you don't do anything with the new instance. To update the `coveringItem` you need to replace it with the new instance.

kaklakariada (changes requested):
* Please update PR description to `Closes #572`
* Please add an entry to the changelog including your name, so you get the acknowledgement for your contribution

api/src/main/java/org/itsallcode/openfasttrace/api/core/LinkedSpecificationItem.java
redcatbear: Thanks for your contribution @bkalika . But the problem lies a lot deeper and needs more refactoring. I have that on my todo list.

The short version is that we need to distinguish between items and links to avoid mutable objects. With links we will have a append only list. I will fix that at some point, but it's not my priority right now.

@kaklakariada, I think we should close this PR and have an issue ticket for the refactoring instead.

Also, this is not a feature PR.

bkalika: @redcatbear, thanks for the clarification! That makes sense. I understand the deeper refactoring should separate specification items from links and make links append-only. If you already have a design in mind for the bigger refactoring, I’m happy to leave it to you. but if you’d like, I can also try to take a first pass at it now while I already have the code and the problem in context.

kaklakariada:
Closing this PR as discussed.

